PR #6, opened 2023-09-01
Keycloak authentication

petri-nykanen:
Login and logout functions are now working.

Additional changes:
-Added config.ts -file to setup keycloak and cleanEnv for validating env-variables.
-Reigned the tyranny of ESLint just a wee lil bit


package.json
```diff
@@ -17,6 +17,8 @@
     "@mui/material": "^5.14.7",
     "@types/react-router": "^5.1.20",
     "@types/react-router-dom": "^5.3.3",
+    "envalid": "^7.3.1",
+    "keycloak-js": "^22.0.1",
```
villejuutila, line 21: ´keycloak-js` versioning follows the versioning of Keycloak itself. e.g. this version is meant for Keycloak v. 22.0.1. Typically they've been backwards compatible but just in case I'd use the correct version. Timebank staging Keycloak seems to be version 19.0.2.
The Keycloaks version should be bumped to the most up to date (22.0.1).

src/components/providers/authentication-provider.tsx
```diff
@@ -0,0 +1,78 @@
...
+  children: JSX.Element
+}
+
+const keycloak = new Keycloak(config.auth);
+/**
+* Provides Keycloak authentication functions, such as login and logout
+*/
+const AuthenticationProvider = ({ auth, setAuth, setUserProfile, children }: Props) => {
+  const updateAuthData = useCallback(() => {
+    console.log(auth);
```
villejuutila, line 20: Remove `console.log`

src/components/screens/error-page.tsx
```diff
@@ -7,16 +7,14 @@ import { useRouteError } from "react-router-dom";
...
 
   return (
     <div id="error-page">
       <h1>Oops!</h1>
       <p>Sorry, an unexpected error has occurred.</p>
       <p>
         <i>
-          {error.statusText}
-          {error.message}
+          {(error as Error)?.message || (error as { statusText?: string })?.statusText}
```
villejuutila, line 17: What's the point of this `error-page`?

src/App.tsx
```diff
@@ -1,23 +1,41 @@
...
+  logout: () => void
+}
 
 /**
  * Application component
  *
  */
 function App () {
+  const [auth, setAuth] = useState<Auth>();
+  const [userProfile, setUserProfile] = useState<KeycloakProfile | undefined>(undefined);
```
villejuutila, line 20:
```suggestion
  const [userProfile, setUserProfile] = useState<KeycloakProfile>();
```
`undefined` value doesn't need to be explicitly typed nor set.

src/App.tsx
```diff
@@ -1,23 +1,41 @@
...
+  tokenRaw: string | undefined
+  logout: () => void
+}
 
 /**
  * Application component
  *
  */
 function App () {
+  const [auth, setAuth] = useState<Auth>();
```
villejuutila, line 19: Even after seeing the discussions taken place in Slack, I'd implement this with Jotai state management since the beginning. Setup is so simple that I just simply don't see any drawbacks in doing that.

.eslintrc.json
```diff
@@ -22,6 +22,9 @@
     "react/react-in-jsx-scope": "off",
     "react/jsx-uses-react": "off",
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx", ".ts", ".tsx"] }],
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/strict-boolean-expressions": "off",
+    "@typescript-eslint/no-floating-promises": "off",
```
villejuutila, line 27: These are all good rules and imo shouldn't be turned off.
Another thing worth noting is that during this year our company has shifted quite a lot from ESLint towards Rome. https://rome.tools. Rome is a new kid on the block with auto-formatter. I'd highly recommend replacing ESLint in this project with Rome aswell.

Reference for that can be found in;
https://github.com/Metatavu/oss-surveys-management/blob/develop/.vscode/settings.json
https://github.com/Metatavu/oss-surveys-management/blob/develop/.vscode/extensions.json
https://github.com/Metatavu/oss-surveys-management/blob/develop/rome.json
https://github.com/Metatavu/oss-surveys-management/blob/25687220b496d089e317d7082bc2661ce29fbfc8/package.json#L48

src/App.tsx
```diff
@@ -1,23 +1,41 @@
 import { RouterProvider, createBrowserRouter } from "react-router-dom";
 import DashboardScreen from "./components/screens/dashboard/dashboard-screen";
 import ErrorPage from "./components/screens/error-page";
+import AuthenticationProvider from "./components/providers/authentication-provider";
+import { useState } from "react";
+import { type KeycloakProfile, type KeycloakTokenParsed } from "keycloak-js";
```
villejuutila, line 6:
```suggestion
import type { KeycloakProfile, KeycloakTokenParsed } from "keycloak-js";
```

villejuutila (changes requested):

src/components/providers/authentication-provider.tsx
```diff
@@ -0,0 +1,78 @@
...
+    setAuth({
+      token: keycloak.tokenParsed!,
+      tokenRaw: keycloak.token!,
```
JuusoMRusanen, line 23: I think we should avoid Non-null Assertions (variables followed with ! operator). You could check these with an IF -statement beforehand.

src/App.tsx
```diff
@@ -1,23 +1,41 @@
...
  *
  */
 function App () {
+  const [auth, setAuth] = useState<Auth>();
+  const [userProfile, setUserProfile] = useState<KeycloakProfile | undefined>(undefined);
+
+  const router = createBrowserRouter([
+    {
+      path: "/",
+      element: <DashboardScreen auth={auth!}/>,
```
JuusoMRusanen, line 25:
```suggestion
      element: <DashboardScreen auth={auth!} />,
```

src/App.tsx
```diff
@@ -1,23 +1,41 @@
...
+
   return (
     <div className="App">
-      <RouterProvider router={router}/>
+      <AuthenticationProvider
+        auth={auth}
+        setAuth={setAuth}
+        userProfile={userProfile}
+        setUserProfile={setUserProfile}>
+        <RouterProvider router={router}/>
```
JuusoMRusanen, line 37:
```suggestion
        <RouterProvider router={router} />
```

JuusoMRusanen (changes requested):
Changes requested

src/components/providers/authentication-provider.tsx
petri-nykanen: Removed console.log

src/components/screens/error-page.tsx
JuusoMRusanen: > What's the point of this `error-page`?
I made an error-page to be an error element because react-router-dom official docs recommended using one. However, I briefly discussed this with Eetu this morning, as to why oss-survey-management app doesn't use one, apparently there's no point in it.

JuusoMRusanen: > What's the point of this `error-page`?

I made an error-page to be an error element because react-router-dom official docs recommended using one. However, I briefly discussed this with Eetu this morning, as to why oss-survey-management app doesn't use one, apparently there's no point in it.

petri-nykanen: For now, the error-page exists to be displayed if the accessed route/path does not exist. Hopefully it also could be configured to displayed if the logged user is unauthorized to access said route. 

We were discussing about the necessity of this page as opposed to using dialog-based error handling for example. I'll leave it as it is for now.

src/App.tsx
petri-nykanen: Fixed

.eslintrc.json
JuusoMRusanen: I would lean towards changing to Rome as well, even though it's my action that ESLint is used right now. We should discuss about this and maybe add a new task.

src/App.tsx
petri-nykanen: Fixed

src/App.tsx
petri-nykanen: Fixed

src/App.tsx
petri-nykanen: Fixed

package.json
petri-nykanen: Fixed

src/components/providers/authentication-provider.tsx
petri-nykanen: Fixed using Rome suggestion
https://docs.rome.tools/lint/rules/nononnullassertion/

src/App.tsx
petri-nykanen: This seems to be a task of its own in the Trello board. Will configure Jotai state management in a different issue and branch.

.eslintrc.json
petri-nykanen: Replaced ESLint with Rome.

JuusoMRusanen (approved):

PR #9, opened 2023-09-04
Feature 3 new error handling

JuusoMRusanen:
Added an error handler wrapper using Jotai, and a generic dialog for it. 
Note: merging this will add Jotai state manager to this project.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,10 @@
 import { RouterProvider, createBrowserRouter } from "react-router-dom";
 import DashboardScreen from "./components/screens/dashboard/dashboard-screen";
-import ErrorPage from "./components/screens/error-page";
+import ErrorScreen from "./components/screens/error-screen";
 import AuthenticationProvider from "./components/providers/authentication-provider";
 import { useState } from "react";
 import type { KeycloakProfile, KeycloakTokenParsed } from "keycloak-js";
+import ErrorHandler from "./components/contexts/error-handler";
 
 export interface Auth {
   token: KeycloakTokenParsed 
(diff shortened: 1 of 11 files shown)


src/components/generics/generic-dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+  onConfirm,
+  error,
+  fullScreen,
+  fullWidth,
+  maxWidth,
+  disableEnforceFocus,
+  disabled,
+  ignoreOutsideClicks,
+  children
+}: Props) => {
```
petri-nykanen, line 55: Dialog component has a list of props, some which are unused. For me this is fine, but if you want to more concise code then maybe the unnecessary props could be dropped for now.

src/App.tsx
```diff
@@ -23,19 +24,21 @@ function App () {
...
-        userProfile={userProfile}
-        setUserProfile={setUserProfile}>
-        <RouterProvider router={router} />
-      </AuthenticationProvider>
+      <ErrorHandler>
+        <AuthenticationProvider
+          auth={auth}
+          setAuth={setAuth}
+          userProfile={userProfile}
+          setUserProfile={setUserProfile}>
```
Jdallos, line 38: If jotai is installed then shouldn't these props be accessed as atoms within the AuthenticationProvider

src/components/contexts/error-handler.tsx
```diff
@@ -0,0 +1,85 @@
...
+const ErrorHandler = ({ children }: Props) => {
+  const [error, setError] = useAtom(errorAtom);
+
+  /**
+   * Handles error message
+   *
+   * @param message error message
+   * @param err any error
+   */
+  const handleError = async (message: string, err?: any) => {
```
Jdallos, line 28: Can we type err with something more specific than any?

src/components/screens/error-screen.tsx
```diff
@@ -5,7 +5,7 @@ import { useRouteError } from "react-router-dom";
  * 
  * @returns ReactElement
  */
-function ErrorPage () {
+function ErrorScreen () {
```
Jdallos, line 8: Convert to arrow function

Jdallos (approved):

src/components/generics/generic-dialog.tsx
JuusoMRusanen: I think these are useful options to have in the generic dialog.

src/App.tsx
JuusoMRusanen: Made the props accessible within the AuthenticationProvider using Jotai.

src/components/contexts/error-handler.tsx
JuusoMRusanen: Typed it as an object

src/components/screens/error-screen.tsx
JuusoMRusanen: done

PR #12, opened 2023-09-05
Featured method to change language of the app

samisii:


JuusoMRusanen (changes requested):
Request changes

src/components/providers/authentication-provider.tsx
```diff
@@ -5,7 +5,7 @@ import Keycloak from "keycloak-js";
 import { useCallback, useEffect } from "react";
 
 interface Props {
-  children: JSX.Element;
+  children: JSX.Element[];
```
petri-nykanen, line 8: What does this array notation mean or change? Not that I know, just curious. Just seeing JSX.Element, ReactElement and ReactNode used interchangably.

Jdallos: Good question, chatGpt provided me with an informative answer!

Jdallos, line 8: Is this change necessary?

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,19 +1,22 @@
-import { useAtom } from "jotai";
-import { authAtom } from "../../../atoms/auth";
+import React from "react";
```
Jdallos, line 1: Avoid importing React as done here, rather use destructured imports e.g. Import { useEffect } from React

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,19 +1,22 @@
...
+import { useLocale } from "../../../localization/LanguageSwitcher";
 
 /**
  * Dashboard screen component
- * 
  */
 const DashboardScreen = () => {
+  const { locale } = useLocale();
   const [auth] = useAtom(authAtom);
+  React.useEffect(() => {
```
Jdallos, line 11: Change the above import then can just use useEffect here

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,19 +1,22 @@
...
+import strings from "../../../localization/strings";
+import { useLocale } from "../../../localization/LanguageSwitcher";
 
 /**
  * Dashboard screen component
- * 
  */
 const DashboardScreen = () => {
+  const { locale } = useLocale();
   const [auth] = useAtom(authAtom);
```
Jdallos, line 10: useAtom is not imported which is causing an error preventing the dashboard-screen rendering

src/localization/strings.ts
```diff
@@ -0,0 +1,24 @@
+import LocalizedStrings, { type LocalizedStringsMethods } from "localized-strings";
```
Jdallos, line 1: type is not needed here

src/localization/strings.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Type representing the shape of your translation strings
+ */
+interface TranslationStrings {
+  notYetImplemented: string
+}
+
+/**
+ * Interface describing localized strings
+ */
+export interface IStrings extends LocalizedStringsMethods, TranslationStrings {
```
Jdallos, line 15: Again this seems a little different to the example in OSS which should be the guide to follow.
e.g.export interface IStrings extends LocalizedStringsMethods, TranslationStrings could be 
export interface Localized extends LocalizedStringsMethods

const strings: IStrings = new LocalizedStrings could be
const strings: Localized = new LocalizedStrings

So really just a difference in naming, but I'd recommend keeping the names the same as they are in other projects,

src/localization/en.json
```diff
@@ -0,0 +1,4 @@
+{
+    "notYetImplemented": "not yet implemented"
+
```
Jdallos, line 3:
```suggestion
```

src/localization/LanguageSwitcher.tsx
```diff
@@ -0,0 +1,79 @@
+import { type ReactNode, createContext, useContext, useState } from "react";
```
Jdallos, line 1: I've not come across a LanguageSwitcher before, is this actually needed I don't see anything like this in OSS

src/localization/fi.json
```diff
@@ -0,0 +1,3 @@
+{
+    "notYetImplemented": "ei vielä toteutettu"
+  }
```
Jdallos, line 3:
```suggestion
}
```

src/types/index.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Type describing available languages
+ */
+export type Language = "fi" | "en";
```
Jdallos, line 4: As far as I can tell this type is unused and so can be removed

tsconfig.json
```diff
@@ -1,21 +1,26 @@
 {
   "compilerOptions": {
     "target": "ES2020",
-    "useDefineForClassFields": true,
     "lib": ["ES2020", "DOM", "DOM.Iterable"],
     "module": "ESNext",
     "skipLibCheck": true,
     "useUnknownInCatchVariables": true,
     "strict": true,
+    "esModuleInterop": true,
```
Jdallos, line 9: Lets discuss the reasons for these config changes and check if they are necessary, I'd be hesitant to make any changes to the config without good reason

Jdallos (changes requested):
Some changes requested, happy to discuss if needed

src/components/providers/authentication-provider.tsx
samisii: Juuso came across with this solution. There was an error for multiple children, and I coudn't resolve it.

src/localization/LanguageSwitcher.tsx
samisii: Didn't realize that we'll reference OSS on this project, sorry. But this was actually just a "project name" for the file, did think in the beginning that it's not a good name for file but somehow ended up keeping it... Will rename appropriately.

src/components/providers/authentication-provider.tsx
JuusoMRusanen: The code started working after this change

Jdallos: I removed it locally and it doesn't seem to cause any problems, also wasn't it working without this previously and it doesn't look like any other changes have being made in this file?

src/localization/LanguageSwitcher.tsx
Jdallos: I could be mistaken but I believe this whole file can be removed and is not necessary, we just use the localization package to achieve what we need. We can check this together though

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,19 +1,27 @@
...
- * 
  */
 const DashboardScreen = () => {
+  const [locale] = useAtom(localeAtom);
+  
   const [auth] = useAtom(authAtom);
 
+  useEffect(() => {
+    strings.setLanguage(locale);
+  }, [locale]);
```
Jdallos, line 17: This works but will become problematic if we need to do this in every component. See the OSS atoms/language.ts you could use this file to replace you r current localeAtom.ts and this will take care of setting the language without having to use the useEffect in this way.

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,19 +1,27 @@
...
+
   return (
-    <div>
-      <div>This is where we would put our dashboard! IF WE HAD ONE!</div>
+    <>
+      <div>{strings.notYetImplemented}</div>
       <button type="button" onClick={ auth?.logout }>Log out</button>
```
Jdallos, line 22: Logout string should be localized

src/types/index.ts
Jdallos, line 1: Empty file/ folder should be removed

tsconfig.json
```diff
@@ -1,21 +1,25 @@
...
-    "allowImportingTsExtensions": true,
+    "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "react-jsx",
   },
-  "include": ["src"],
-  "references": [{ "path": "./tsconfig.node.json" }]
+  "include": ["src/**/*.ts", "src/**/*.tsx", "localization/**/*.ts"],
```
Jdallos, line 18: As said in the previous comment, I don't think these changes to the config are necessary, revert this file so that is matches what is currently in develop.

Jdallos (changes requested):
Changes requested

src/App.tsx
```diff
@@ -20,11 +22,14 @@ function App () {
 
   return (
     <div className="App">
-      <ErrorHandler>
-        <AuthenticationProvider>
-          <RouterProvider router={router} />
-        </AuthenticationProvider>
-      </ErrorHandler>
+      <Provider>
```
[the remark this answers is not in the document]
samisii: You're right, it didn't do anything. LocaleProvider confused me a little. Deleted now

src/components/screens/dashboard/dashboard-screen.tsx
samisii: Modified dashboard to use languageAtom

src/App.tsx
```diff
@@ -23,12 +23,14 @@ function App () {
 
   return (
     <div className="App">
-        <ErrorHandler>
+      <ThemeProvider theme={theme}>
+      <ErrorHandler>
```
joonas-ris, line 27: Fix indent

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,24 +1,63 @@
...
+ */
+  const getPersons = async () => {
+    setIsLoading(true);
+    const fetchedPersons = await personsApi.listPersons({ active: true });
+    const loggedInPerson = fetchedPersons.filter(person => person.keycloakId === userProfile?.id);
+
+    if (loggedInPerson.length) {
+      try {
+          const fetchedPerson = await personsApi.listPersonTotalTime({personId: loggedInPerson[0].id});
+          setPersonTotalTime(fetchedPerson[0]);
```
joonas-ris, line 35: Fix indent

src/components/locale-toggle/LocaleToggle.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { languageAtom } from '../../atoms/languageAtom';
+import { Language } from '../../types';
+
+/**
+ * Function to change language of the app
+ */
+function LocaleToggle() {
+    const [language, setLanguage] = useAtom(languageAtom);
+
+    const handleLocaleChange = (newLanguage: string) => {
```
Jdallos, line 13: Missing JSDoc comments

src/components/locale-toggle/LocaleToggle.tsx
```diff
@@ -0,0 +1,44 @@
...
+        setLanguage(newLanguage as Language);
+        strings.setLanguage(newLanguage);
```
Jdallos, line 15: I think this is already being taken care of within the languageAtom changes you have made, so this line strings.setLanguage(newLanguage);
Can be removed here

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -17,6 +19,7 @@ function DashboardScreen () {
   const [isLoading, setIsLoading] = useState(false);
   const [personTotalTime, setPersonTotalTime] = useState<PersonTotalTime>();
   const { personsApi } = useApi();
+  const [_language] = useAtom(languageAtom);
```
Jdallos, line 22: It looks like this is unused and marked as so, so can this just be removed?

src/localization/en.json
```diff
@@ -0,0 +1,9 @@
+{
```
Jdallos, line 1: As the app grows we will want to structure this more, for example could logout and currentLocale strings be in a header object?

src/localization/fi.json
```diff
@@ -0,0 +1,9 @@
+{
+  "notYetImplemented": "Ei vielä toteutettu",
```
Jdallos, line 2: Update structure as said in comment for en.json

src/localization/strings.ts
```diff
@@ -0,0 +1,36 @@
...
+  fi: string;
+}
+
+/**
+* Translations interface
+*/
+interface Translations {
+  notYetImplemented: string;
+  currentLocaleLabel: string;
+  localization: Localizations;
```
Jdallos, line 19: Typically the localization would be defined in the main interface Translations (which is typically called Localized) rather than as a separate interface.

src/localization/strings.ts
```diff
@@ -0,0 +1,36 @@
...
+  logout: string;
+}
+
+/**
+* Localized strings
+*/
+export interface Localized extends LocalizedStringsMethods, Translations {}
```
Jdallos, line 26: I don't think there is any advantage to doing it this way. Typically the interface you have called Translations would be called Localized and extend LocalizedStringsMethods, rather than doing it here as you have done. Let me know if there is a reason for doing it this way I have missed

tsconfig.json
Jdallos, line 1: If there are no changes to this file you can right click on the file in the vsCode source control section and discard all changes for it.

Jdallos (changes requested):

src/components/providers/authentication-provider.tsx
JuusoMRusanen: Indeed does work without this change. Reversed the change.

JuusoMRusanen:
Made changes on behalf of Sami

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -6,53 +6,58 @@ import { PersonTotalTime } from "../../../generated/client";
 import { errorAtom } from "../../../atoms/error";
 import LoaderWrapper from "../../generics/loader-wrapper";
 import { getHoursAndMinutes } from "../../../utils/time-utils";
+import strings from "../../../localization/strings";
 
 /**
  * Dashboard screen component
  */
-function DashboardScreen () {
+function DashboardScreen() {
```
Jdallos, line 14: Components should be declared using arrow function syntax.

Jdallos (changes requested):

Jdallos (changes requested):

package.json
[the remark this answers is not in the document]
JuusoMRusanen: I removed all the unused dependencies and nothing broke.

src/App.tsx
```diff
@@ -10,8 +11,7 @@ import { theme } from "./theme";
  * Application component
  *
  */
-function App () {
-
+function App() {
```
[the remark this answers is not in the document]
JuusoMRusanen: fixed

src/components/localization-buttons/LocalizationButtons.tsx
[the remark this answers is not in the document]
JuusoMRusanen: Renamed folder to layout-components.

src/components/localization-buttons/LocalizationButtons.tsx
```diff
@@ -0,0 +1,47 @@
+import { useAtom } from "jotai";
+import strings from "../../localization/strings";
+import { Button } from "@mui/material";
+import { languageAtom } from "../../atoms/languageAtom";
+import { Language } from "../../types";
+
+/**
+ * LocalizationButtons component
+ */
+const LocalizationButtons: React.FC = () => {
```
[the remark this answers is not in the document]
JuusoMRusanen: Removed React.FC

src/components/screens/dashboard/dashboard-screen.tsx
JuusoMRusanen: Changed to arrow function

src/components/localization-buttons/LocalizationButtons.tsx
```diff
@@ -0,0 +1,47 @@
...
+      ))}
+    </div>
+  );
+
+  return (
+    <div>
+      {renderLocalizationButtons()}
+      <p>
+        {strings.label.currentLocaleLabel} {language}
+      </p>
```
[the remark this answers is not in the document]
JuusoMRusanen: Came up with a different solution: Changed Button component to ButtonToggle component, which shows which localization is selected by highlighting the corresponding button.

Jdallos (changes requested):

src/components/layout-components/localization-buttons.tsx
```diff
@@ -0,0 +1,49 @@
...
+    newLanguage: Language | null
+  ) => {
+    setLanguage(newLanguage as Language);
+  };
+
+  /**
+   * Renders localization buttons
+   */
+  const renderLocalizationButtons = () => (
+    <div>
```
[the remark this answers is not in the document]
JuusoMRusanen: Nope. Removed it.

src/components/layout-components/localization-buttons.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * LocalizationButtons component
+ */
+const LocalizationButtons = () => {
+  const [language, setLanguage] = useAtom(languageAtom);
+
+  /**
+   * Method to handle locale change
+   */
+  const handleLocaleChange = (
+    event: React.MouseEvent<HTMLElement>,
```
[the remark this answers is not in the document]
JuusoMRusanen: Good advice. Fixed.

src/components/screens/dashboard/dashboard-screen.tsx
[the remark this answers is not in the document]
JuusoMRusanen: Localized 'Your balance' string and fixed localization toggle.

src/components/layout-components/localization-buttons.tsx
```diff
@@ -13,34 +14,29 @@ const LocalizationButtons = () => {
   /**
    * Method to handle locale change
    */
-  const handleLocaleChange = (
-    event: React.MouseEvent<HTMLElement>,
-    newLanguage: Language | null
-  ) => {
+  const handleLocaleChange = (_event: MouseEvent<HTMLElement>, newLanguage: Language | null) => {
```
Jdallos, line 17: Is there a reason newLanguage is typed as Language | null?
If so, it might be a good idea to check if the newLanguage is null and handle appropriately rather than end up setting the localization to null.

src/components/layout-components/localization-buttons.tsx
```diff
@@ -13,34 +14,29 @@ const LocalizationButtons = () => {
...
-    </div>
+    <ToggleButtonGroup
+      value={language}
+      exclusive
+      onChange={handleLocaleChange}
+      aria-label="localization"
+    >
+      {strings.getAvailableLanguages().map((availableLanguage, idx) => {
+        const value = strings.getString(`localization.${availableLanguage}`, language);
+        return (
```
Jdallos, line 33: Having a line break before a return value is a good general rule for code readability

Jdallos (approved):
Approved with a couple of comments

src/components/layout-components/localization-buttons.tsx
JuusoMRusanen: Removed null and added null check.

src/components/layout-components/localization-buttons.tsx
JuusoMRusanen: Added a line break.

PR #13, opened 2023-09-05
Feature 10 timebank api

petri-nykanen:
Api has been configured to fetch person data and balance.


src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,62 @@
...
+    const loggedInPerson = await (await personsApi.listPersons({ active: true })).filter(person => person.keycloakId === userProfile?.id)
+
+    if (loggedInPerson.length > 0) {
+      try {
+        if (auth?.tokenRaw){
+          const fetchedPerson : PersonTotalTime[] = await personsApi.listPersonTotalTime({personId: loggedInPerson[0].id});
+          setPersonTotalTime(fetchedPerson);
+        }
+      } catch (error) {
+        console.error(error);
```
JuusoMRusanen, line 35: We have an error handler to use now :), should set all the errors to errorAtom.
For example:
`const [ error, setError ] = useAtom(errorAtom); // This line is for upper scope`
`setError(`${ "Person fetch has failed." }, ${ error }`);`

JuusoMRusanen (changes requested):
Request changes

src/components/screens/dashboard/dashboard-screen.tsx
petri-nykanen: Fixed

JuusoMRusanen (approved):

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,61 @@
...
+  
 
   return (
-    <div>
-      <div>This is where we would put our dashboard! IF WE HAD ONE!</div>
-      <button type="button" onClick={ auth?.logout }>Log out</button>
-    </div>
+    <>{(isLoading)
+      ?<div>Loading...</div>
+      :<div>
```
Jdallos, line 53: Use a loaderWrapper instead of this as done in OSS

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,61 @@
...
+        <div>{personTotalTime?.map((person) => {
+          return (`Your balance is ${Math.trunc(person.balance / 60)} h ${(person.balance % 60) * -1} min`)
```
Jdallos, line 55: I had a look at the original time bank and this balance logic seems a little different, so I assume you are sure its generating the correct balances (or at least the same as the original time bank was)? Perhaps this balance calculation could be put in its own function which returns the balance string as I think this will probably be reused somewhere

Jdallos (changes requested):
Some changes requested

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,62 @@
...
  * 
  */
-const DashboardScreen = () => {
-  const [auth] = useAtom(authAtom);
+function DashboardScreen () {
+  const auth = useAtomValue(authAtom)
+  const userProfile = useAtomValue(userProfileAtom);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [errorMsg, setErrorMsg] = useState<string>();
+  const [personTotalTime, setPersonTotalTime] = useState<PersonTotalTime[]>();
```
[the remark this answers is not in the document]
Jdallos: I think if you do this the type of personTotalTime will be PersonTotalTime rather than PersonTotalTime | undefined

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,62 @@
...
 /**
  * Dashboard screen component
  * 
  */
-const DashboardScreen = () => {
-  const [auth] = useAtom(authAtom);
+function DashboardScreen () {
+  const auth = useAtomValue(authAtom)
+  const userProfile = useAtomValue(userProfileAtom);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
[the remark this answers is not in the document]
petri-nykanen: Fixed

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,62 @@
...
+  const [errorMsg, setErrorMsg] = useState<string>();
+  const [personTotalTime, setPersonTotalTime] = useState<PersonTotalTime[]>();
+  const { personsApi } = useApi();
+  
+/**
+ * Initialize logged in person's time data.
+ *
```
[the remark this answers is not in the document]
petri-nykanen: Fixed

src/components/screens/dashboard/dashboard-screen.tsx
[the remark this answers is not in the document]
petri-nykanen: Adding empty array to personTotalTime will indeed change its typing, therefore removing the "?" operators from personTotalTime expressions.

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,62 @@
...
+  const [personTotalTime, setPersonTotalTime] = useState<PersonTotalTime[]>();
+  const { personsApi } = useApi();
+  
+/**
+ * Initialize logged in person's time data.
+ * 
+ */
+  const getPersons = async () => {
+    setIsLoading(true);
+    //NOTE: Trainees do not have time bank entries, therefore a use keycloak ID from another employee for testing.
```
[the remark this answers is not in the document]
petri-nykanen: Comment removed

src/components/screens/dashboard/dashboard-screen.tsx
petri-nykanen: Compared the time values to the ones in the previous time bank and they seemed identical

src/components/generics/loader-wrapper.tsx
```diff
@@ -0,0 +1,28 @@
+import { Backdrop, CircularProgress } from "@mui/material";
+
+/**
+ * Components properties
+ */
+interface Props {
+  loading: boolean;
+  children: React.ReactNode;
```
Jdallos, line 8: Use destructured imports e.g. import {ReactNode} from "react", rather than React.ReactNode

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,57 @@
...
- * 
  */
-const DashboardScreen = () => {
-  const [auth] = useAtom(authAtom);
+function DashboardScreen () {
+  const auth = useAtomValue(authAtom);
+  const userProfile = useAtomValue(userProfileAtom);
+  const setError = useSetAtom(errorAtom);
+  const [isLoading, setIsLoading] = useState(false);
+  const [personTotalTime, setPersonTotalTime] = useState<PersonTotalTime[]>([]);
```
Jdallos, line 18: Does personTotalTime need to be an array, if not this could be changed and you would not need to map in the component return (which could be a bit confusing if there is only ever one PersonTotalTime in this case.

src/components/screens/dashboard/dashboard-screen.tsx
```diff
@@ -1,18 +1,57 @@
...
 
   return (
-    <div>
-      <div>This is where we would put our dashboard! IF WE HAD ONE!</div>
-      <button type="button" onClick={ auth?.logout }>Log out</button>
-    </div>
+    <LoaderWrapper loading={isLoading}>
+        <div>{personTotalTime.map((person) => {
+          return (`Your balance is ${getHoursAndMinutes(person.balance)}`)
+      })}</div>
```
Jdallos, line 52: Clean up the indentation here

src/utils/time-utils.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Converts inputted minutes into hours and minutes
```
Jdallos, line 2: @param comment needed

Jdallos (approved):
Great, few changes requested and once these are done its approved

PR #16, opened 2023-09-05
Feature 15 set up master theme

JuusoMRusanen:
Request review

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,7 @@ import DashboardScreen from "./components/screens/dashboard/dashboard-screen";
 import ErrorScreen from "./components/screens/error-screen";
 import AuthenticationProvider from "./components/providers/authentication-provider";
 import ErrorHandler from "./components/contexts/error-handler";
-import { ThemeProvider } from "@emotion/react";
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -3,6 +3,8 @@ import DashboardScreen from "./components/screens/dashboard/dashboard-screen";
 import ErrorScreen from "./components/screens/error-screen";
 import AuthenticationProvider from "./components/providers/authentication-provider";
 import ErrorHandler from "./components/contexts/error-handler";
+import { ThemeProvider } from "@emotion/react";
```
Jdallos, line 6: This looks like a different import from what is used in OSS, is that deliberate? I'm not familiar with the details for setting up a theme, so if it works and you're happy with it then so am I!

Jdallos (approved):
Approved with a comment

src/App.tsx
JuusoMRusanen: According to the docs, createTheme is imported from "@mui/material". Thanks for pointing that out!

petri-nykanen (approved):
Approved

PR #20, opened 2023-09-06
Make npm run dev command open the browser automatically

JuusoMRusanen:
Request review

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -4,5 +4,7 @@ import react from '@vitejs/plugin-react'
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [react()],
-
-})
+  server: {
+    open: true,
+  },
+});


joonas-ris (approved):

vite.config.ts
```diff
@@ -4,5 +4,7 @@ import react from '@vitejs/plugin-react'
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [react()],
-
-})
+  server: {
+    open: true,
```
Jdallos, line 8: Strange that this isn't needed in the OSS project to achieve the same behaviour, perhaps its due to this project using a newer version of vite...

Jdallos (approved):

vite.config.ts
JuusoMRusanen: Perhaps

PR #59, opened 2023-11-29
Feature 52 fix bad language

sipuli7:
Modified some displayed language.


JuusoMRusanen (approved):
I guess we cannot get rid of the maternityPaternityLeave key unless we change it from the API Spec.

Jdallos (approved):

PR #63, opened 2023-12-12
Bugfix 63 timeperiod invalid date

JuusoMRusanen:
Fix timebank screen timeperiod displaying "InvalidDate" on Safari browser by switching to Luxon date formatting

### DIFF
--- a/src/utils/time-utils.ts
+++ b/src/utils/time-utils.ts
@@ -1,5 +1,23 @@
 import { DateTime, Duration } from "luxon";
-import strings from "../localization/strings";
+import { useAtomValue } from "jotai";
+import { languageAtom } from "../atoms/language";
+
+/**
+ * Format date
+ *
+ * @param date datetime object
+ * @param dateWithTime datetime object with time
+ * @returns formatted date time
+ */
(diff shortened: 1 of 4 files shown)


src/utils/time-utils.ts
```diff
@@ -30,14 +48,15 @@ export const getHours = (minutes: number) =>
...
 
   if (timespan[0].length > 4) {
-    const startDate = new Date(timespan[0].split("-").join(", ")).toLocaleDateString(
-      strings.localization.time
-    );
-    const endDate = new Date(timespan[1].split("-").join(", ")).toLocaleDateString(
-      strings.localization.time
-    );
+    const startDate = DateTime.fromJSDate(new Date(timespan[0]))
+      .setLocale(language)
```
villejuutila, line 55: Isn't Luxons locale already defined in `App.tsx`

villejuutila (approved):

src/utils/time-utils.ts
JuusoMRusanen: Yes, it indeed is. Removed the set locales.

PR #66, opened 2023-12-19
Feature 37 timebank overview range picker

JuusoMRusanen:


Jdallos:
This branch was worked on by multiple interns and became suspiciously bloated, rather than invest more time in this low priority feature, the PR is closed and if the feature is required in the future the task can be started again.

PR #68, opened 2023-12-22
Feature 60 vacation days

sipuli7:
Feature current users vacation days on the main page balance card and bottom of the vacations screen.


src/components/home/vacations-card.tsx
```diff
@@ -143,11 +149,59 @@ const VacationsCard = () => {
     fetchVacationsRequests();
   }, []);
 
+  /**
+   * Initialize logged in person's data.
+   */
+  const getPersonVacations = async () => {
```
JuusoMRusanen, line 155: This function name could be reduced to "getPersons", as this is what the function seems to do.

src/components/home/vacations-card.tsx
```diff
@@ -143,11 +149,59 @@ const VacationsCard = () => {
...
+        setPersons(fetchedPersons);
+      }
+    }
+  };
+
+  useMemo(() => {
+    if (!persons) {
+      getPersonVacations();
+    }
+  }, [persons]);
```
JuusoMRusanen, line 170: On row 156 you check for persons.length and on row 167 you check for !persons. These checks are in conflict. I think you meant to check for !persons.length in the function. Please test and verify if this is the case.

src/components/home/vacations-card.tsx
```diff
@@ -143,11 +149,59 @@ const VacationsCard = () => {
...
+
+  /**
+   * display persons vacation days
+   */
+  const renderVacations = (person: Person | undefined) => {
```
JuusoMRusanen, line 175: Update JSDocs and maybe rename function to renderVacationDays, like it's described in the docs. Person parameter is used in the function and @param is missing from the docs.

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -39,6 +43,10 @@ const VacationRequestsScreen = () => {
     adminMode ? allVacationRequestStatusesAtom : vacationRequestStatusesAtom
   );
   const [loading, setLoading] = useState(false);
+  const persons = useAtomValue(personsAtom);
+  const loggedInPerson = persons.find((person: Person) => person.keycloakId === userProfile?.id);
+  const { personsApi } = useApi();
+  const setPersons = useSetAtom(personsAtom);
```
JuusoMRusanen, line 49: persons and setPersons could be combined into one tuple with useAtom()

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -73,6 +81,58 @@ const VacationRequestsScreen = () => {
...
+        setPersons(fetchedPersons);
+      }
+    }
+  };
+
+  useMemo(() => {
+    if (!persons) {
+      getPersonVacations();
+    }
+  }, [persons]);
```
JuusoMRusanen, line 102: Apply the same suggestions as in the vacations-card.tsx

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -73,6 +81,58 @@ const VacationRequestsScreen = () => {
...
+
+  /**
+   * Display persons vacation days
+   */
+  const renderVacations = (person: Person | undefined) => {
```
JuusoMRusanen, line 107: Apply the same suggestions as in the vacations-card.tsx

JuusoMRusanen (approved):
Approved with comments

src/components/home/vacations-card.tsx
sipuli7: removed second check.

src/components/screens/vacation-requests-screen.tsx
sipuli7: Combined.

src/components/home/vacations-card.tsx
sipuli7: Changed the functions name.

src/components/home/vacations-card.tsx
sipuli7: Added @param person to docs.

src/components/screens/vacation-requests-screen.tsx
sipuli7: Added.

src/components/screens/vacation-requests-screen.tsx
sipuli7: Added @param person here aswell.

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -210,22 +268,19 @@ const VacationRequestsScreen = () => {
...
-            status: newStatus,
-            message: LocalizationUtils.getLocalizedVacationRequestStatus(newStatus),
-            createdAt: new Date(),
-            createdBy: userProfile.id,
-            updatedAt: new Date(),
-            updatedBy: userProfile.id
-          }
-        });
-
+      const createdVacationRequestStatus = await vacationRequestStatusApi.createVacationRequestStatus({
```
Jdallos, line 271: In this case have a new line after the = so the await is on the same line as the request. IMO doing it this was is clearer

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -424,6 +476,9 @@ const VacationRequestsScreen = () => {
           updateVacationRequestStatuses={updateVacationRequestStatuses}
           loading={loading}
         />
+        {renderVacationDays(
+          persons.find((person) => person.id === loggedInPerson?.id || config.person.id)
```
Jdallos, line 480: As mentioned in the vacations card could you not use the loggedInPerson here rather than the fine. Also some padding or similar needs to be added to the content returned from renderVacationDays here as it looks a little tightly packed in at the bottom of the table.

src/localization/en.json
```diff
@@ -140,7 +140,9 @@
     "vacationType": "Vacation Type:",
     "applicant": "Applicant:",
     "timeOfVacation": "Time of Vacation:",
-    "status": "Status:"
+    "status": "Status:",
+    "spentVacations": "Spent vacation days",
+    "unspentVacations": "Unspent vacation days"
```
Jdallos, line 145: You use colons : for the finnish translation so lets do the same for the english. Also use of a space after the colon is fine rather than the styling option as mentioned in a previous comment. Either way is fine but lets be consistent

Jdallos (changes requested):
Changes requested

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -73,6 +80,57 @@ const VacationRequestsScreen = () => {
     fetchVacationRequestStatuses();
   }, [vacationRequests]);
 
+    /**
+   * Initialize logged in person data.
+   */
+  const getPersons = async () => {
```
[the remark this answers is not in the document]
JuusoMRusanen: Checking for !persons.length would be appropriate here. Note that Persons atom is already set in authentication-provider.tsx on row 83. Seems to work fine without this function, even if the user navigated directly to this view. Unless there is a reason for re-fetching the persons, like when the vacation days get recalculated?

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -73,6 +80,57 @@ const VacationRequestsScreen = () => {
...
+          <Box>
+            <Typography>
+              {strings.vacationsCard.unspentVacations}
+              <span style={{ color: unspentVacationsColor }}>{person.unspentVacations}</span>
+            </Typography>
+          </Box>
+        </Box>
+      );
+    }
+  };
```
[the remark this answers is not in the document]
JuusoMRusanen: I agree. This could be made into it's own component as the only differences are related to the style of the returned parent component, the flex box. Remove the sx property from line 117 and overwrite component on line 116 to a React fragment component (which is this <></>), and then the imported component `renderVacationDays()` could be used like this, for example:

In vacation-requests-screen:
`<Box sx={{ display: "flex", justifyContent: "space-around" }}>renderVacationDays(...)</Box>`

In vacations-card:
`<Box sx={{ display: "flex", flexDirection: "column" }}>renderVacationDays(...)</Box>`

src/localization/en.json
sipuli7: Silly mistake, supposed to have a colon + space. Fixed now.

src/components/home/vacations-card.tsx
```diff
@@ -143,10 +148,59 @@ const VacationsCard = () => {
...
+      person && person.spentVacations > 0 ? theme.palette.success.main : theme.palette.error.main;
+
+    const unspentVacationsColor =
+      person && person.unspentVacations > 0 ? theme.palette.success.main : theme.palette.error.main;
+
+    if (!person && !loading) {
+      return <Typography>{strings.error.fetchFailedNoEntriesGeneral}</Typography>;
+    } else if (person) {
+      return (
+        <Box sx={{ display: "flex", flexDirection: "column" }}>
```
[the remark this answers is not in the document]
sipuli7: Added some margin between them.

src/components/home/vacations-card.tsx
```diff
@@ -368,6 +422,9 @@ const VacationsCard = () => {
             {adminMode ? strings.tableToolbar.manageRequests : strings.tableToolbar.myRequests}
           </h3>
           <Grid container>
+            {renderVacationDays(
+              persons.find((person) => person.id === loggedInPerson?.id || config.person.id)
+            )}
```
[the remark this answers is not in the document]
sipuli7: Added config.person.id to loggedInPerson and used only loggedInPerson here.

src/components/screens/vacation-requests-screen.tsx
sipuli7: Added new line to make it clearer.

src/components/screens/vacation-requests-screen.tsx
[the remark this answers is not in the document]
sipuli7: Created new util, vacation-days-utils.

src/components/home/vacations-card.tsx
```diff
@@ -28,6 +29,9 @@ import { getVacationRequestPersonFullName } from "../../utils/vacation-request-u
 import { validateValueIsNotUndefinedNorNull } from "../../utils/check-utils";
 import { VacationInfoListItem } from "../../types";
 import { formatDate } from "../../utils/time-utils";
+import config from "../../app/config";
+import { renderVacationDays } from "../../utils/vacation-days-utils";
+
 
 /**
```
JuusoMRusanen, line 36:
```suggestion
import { renderVacationDays } from "../../utils/vacation-days-utils";

/**
```

src/components/home/vacations-card.tsx
```diff
@@ -143,6 +151,24 @@ const VacationsCard = () => {
...
+    }
+  };
+
+  useMemo(() => {
+    getPersons();
+  }, [persons]);
+
+
+
   /**
```
JuusoMRusanen, line 172:
```suggestion
  }, [persons]);

  /**
```

src/utils/vacation-days-utils.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Person } from "../generated/client";
+import strings from "../localization/strings";
+import { useState } from "react";
+import { personsAtom } from "../atoms/person";
+import { useAtom } from "jotai";
+
+/**
+ * Display persons vacation days
+ * @param Person timebank person
+ */
```
JuusoMRusanen, line 12:
```suggestion
/**
 * Display persons vacation days
 *
 * @param Person timebank person
 */
```

src/utils/vacation-days-utils.tsx
```diff
@@ -0,0 +1,51 @@
...
+export const renderVacationDays = (person: Person | undefined) => {
```
JuusoMRusanen, line 13: If you remove the undefined type from the person parameter, you could avoid the necessity of checking for the person variable in the rest of the code in this function.

JuusoMRusanen (approved):
Some extra line breaks to remove, and one suggestion.

src/utils/vacation-days-utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { useState } from "react";
+import { personsAtom } from "../atoms/person";
+import { useAtom } from "jotai";
+
+/**
+ * Display persons vacation days
+ * @param Person timebank person
+ */
+export const renderVacationDays = (person: Person) => {
+const loading = useState(true);
```
Jdallos, line 14: I don't think this is correct. If the state is not being changed then it could just be a regular variable rather than state, also as it is not being changed this means the condition below if (!loading && persons.length) doesn't really make sense. Do you still need loading here at all or can it be handled in the parent component e.g. vacations screen and card, where this render function is being called

src/utils/vacation-days-utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+            {strings.vacationsCard.spentVacations}
+            <span style={{ color: spentVacationsColor }}>{person.spentVacations}</span>
+        </Box>
+        <Box sx={{mb: 2}}>
+            {strings.vacationsCard.unspentVacations}
+            <span style={{ color: unspentVacationsColor }}>{person.unspentVacations}</span>
+        </Box>
+        </>
+        );
+    }
```
Jdallos, line 42: Fix indentation

src/utils/vacation-days-utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+        ? theme.palette.success.main
+        : theme.palette.error.main;
+
+const unspentVacationsColor =
+    person.unspentVacations > 0
+        ? theme.palette.success.main
+        : theme.palette.error.main;
+
+if (!loading && persons.length) {
+    return <Typography>{strings.error.fetchFailedNoEntriesGeneral}</Typography>;
```
Jdallos, line 28: Is this the correct error message to use here, wouldn't the error be related to persons rather than just a general no entries?

src/utils/vacation-days-utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+import strings from "../localization/strings";
+import { useState } from "react";
+import { personsAtom } from "../atoms/person";
+import { useAtom } from "jotai";
+
+/**
+ * Display persons vacation days
+ * @param Person timebank person
+ */
+export const renderVacationDays = (person: Person) => {
```
Jdallos, line 13: I think it makes more sense for this to be a component perhaps under components/ generics or a new folder componenets/vacations as it is actually is rendering something. Utils are utilities that usually deal with logic rather than rendering something.

Jdallos (changes requested):

src/utils/vacation-days-utils.tsx
sipuli7: Moved to components/generics and renamed to vacation-days.tsx

src/utils/vacation-days-utils.tsx
sipuli7: Made loading into a regular variable. If it's complitly removed, it fails to fetch the persons data.

src/components/home/vacations-card.tsx
```diff
@@ -143,6 +150,22 @@ const VacationsCard = () => {
...
+      setLoading(true);
+      if (loggedInPerson || config.person.id) setLoading(false);
+      const fetchedPersons = await personsApi.listPersons({ active: true });
+      setPersons(fetchedPersons);
+    }
+  };
+
+  useMemo(() => {
+    getPersons();
+  }, [persons]);
```
Jdallos, line 167: This wasn't used here before and yet the component was still able to access persons from the atom. It doesn't seem like this is needed to me, can it be removed? (Seems to work just fine when it is).

src/components/screens/timebank-screen.tsx
```diff
@@ -43,7 +43,7 @@ const TimebankScreen = () => {
    * Gets person's total time data.
    */
   const getPersonTotalTime = async () => {
-    if (persons.length) {
+    if (!persons.length) {
```
Jdallos, line 46: Again the ! should have only been added to the function used in the  vacations card, using it here prevents the data being fetched for the timebank screen so the ! needs to be removed here.

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -73,6 +82,22 @@ const VacationRequestsScreen = () => {
...
+      if (loggedInPerson || config.person.id) setLoading(false);
+      const fetchedPersons = await personsApi.listPersons({ active: true });
+      setPersons(fetchedPersons);
+    }
+  };
+
+  useMemo(() => {
+    getPersons();
+  }, [persons]);
+
```
Jdallos, line 100: As you said in the now closed comment it looks like this can be removed, I agree, lets remove it

Jdallos (changes requested):
Changes requested

src/components/home/vacations-card.tsx
sipuli7: Removed

src/components/generics/vacation-days.tsx
Jdallos, line 1: Everything seems fine, but as the back end calculation of the vacation days is incorrect I think for now we should comment out this component and either return null or undefined or also comment out the lines where the component is used. You should also leave a TODO: comment explaining why the code is commented out and that it can be reintroduced once the backend vacations calculation has been updated

Jdallos (changes requested):

Jdallos (approved):

PR #78, opened 2024-01-24
Bugfix 76  Pie chart labels

### LINKED ISSUE #76: Pie chart labels bug

DZotoff:
#76 Looking for the possible solution for labelLine feature, which is connect the label and the Pie.

### DIFF
--- a/src/components/charts/timebank-piechart.tsx
+++ b/src/components/charts/timebank-piechart.tsx
@@ -40,26 +40,15 @@ const TimebankPieChart = ({ personDailyEntry }: Props) => {
           cx="50%"
           cy="50%"
           outerRadius={50}
-
           label={(entry) => {
-          if (entry.dataKey == 0) {
-            return null; // do not render any label if dataKey is 0
-          } else if (entry.dataKey !== "") {
-            // rendering for non-empty dataKey
-            return renderCustomizedLabel(entry);
-          }
-        }}
-        
-        labelLine={false}
-          /* 
-          labelLine={(entry) => {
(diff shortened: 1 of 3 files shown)


yuusufisse (approved):
looks good

src/components/charts/timebank-piechart.tsx
```diff
@@ -52,4 +71,5 @@ const TimebankPieChart = ({ personDailyEntry }: Props) => {
   );
 };
 
+
```
Jdallos, line 74: No need for a double line break here

src/components/charts/timebank-piechart.tsx
```diff
@@ -40,8 +40,16 @@ const TimebankPieChart = ({ personDailyEntry }: Props) => {
           cx="50%"
           cy="50%"
           outerRadius={50}
-          label={renderCustomizedLabel}
-        >
+          label={(entry) => {
+            if (!entry.dataKey) {
+              return null; // do not render any label if dataKey is 0 or empty
```
Jdallos, line 45: Although sometimes its good to leave comments in the code if what is happening is unclear, I would say in this case these comments are not needed and can be removed.

src/components/charts/timebank-piechart.tsx
```diff
@@ -40,8 +40,16 @@ const TimebankPieChart = ({ personDailyEntry }: Props) => {
...
+            } else {
+              // rendering for non-empty dataKey
```
Jdallos, line 47: Same here

Jdallos (approved):

PR #82, opened 2024-01-30
loading spinner feature & padding problem

DZotoff:
spinner added, differences in margin/ padding for timeperiod and logged cards fixed

### DIFF
--- a/src/components/timebank/timebank-content.tsx
+++ b/src/components/timebank/timebank-content.tsx
@@ -163,11 +163,19 @@ const TimebankContent = ({ handleDailyEntryChange, loading }: Props) => {
    * @returns JSX.Element consisting of either chart component
    */
   const renderDailyEntryOrRangeChart = () => {
+    if (loading){
+    return <CircularProgress sx={{ margin: "auto",mt: "5%",mb: "5%" }}  /> 
+    }       
     if (byRange.dailyEntries && selectedEntries) {
       return <TimebankMultiBarChart selectedEntries={selectedEntries} />;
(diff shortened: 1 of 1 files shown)


yuusufisse (approved):
looks good to me

src/components/timebank/timebank-content.tsx
```diff
@@ -163,11 +163,19 @@ const TimebankContent = ({ handleDailyEntryChange, loading }: Props) => {
...
+    }       
     if (byRange.dailyEntries && selectedEntries) {
       return <TimebankMultiBarChart selectedEntries={selectedEntries} />;
     }
     if (personDailyEntry) {
-      return <TimebankPieChart personDailyEntry={personDailyEntry} />;
+      return (
+        <>
+        <TimebankPieChart personDailyEntry={personDailyEntry} /> 
+        {renderTimeEntriesList()}
```
Jdallos, line 176: I see why you have moved this renderTimeEntriesList into here, it does relate to the timebankpiechart and so seems like a good idea. It does also need to be moved to be rendered in the above condition also otherwise the TimebankMultiChart is missing the timeentry types.
Also can you think of a more suitable name for this renderTimeEntriesList method - from a quick look it seems to me that it is rendering the time entries types, perhaps renderTimeEntryTypesList?

Jdallos, line 176: Fix indentation

Jdallos (changes requested):
Couple of changes requested

PR #91, opened 2024-02-16
Disabled possibility for user to edit an approved vacation requests.

DZotoff:


src/components/vacation-requests-table/vacation-requests-table-toolbar/vacation-requests-table-toolbar.tsx
```diff
@@ -132,7 +133,19 @@ const TableToolbar = ({
...
+          }
+          {selectedRowIds?.length === 1 && 
+          rows.find((request: VacationsDataGridRow) => request.id === selectedRowIds[0])?.status === VacationRequestStatuses.APPROVED &&
+            <ToolbarGridItem sm={adminMode ? 3 : 6} item xs={6}>
+              <FormToggleButton
+                title={strings.tableToolbar.edit}
+                ButtonIcon={Edit}
+                value={formOpen}
+                setValue={setFormOpen}               
+                disabled={adminMode ? false : true}
```
Jdallos, line 145: I don't think we wanted anyone (including admins) to be able to edit the vacation request after status change, so why isn't this just true?

src/components/vacation-requests-table/vacation-requests-table-toolbar/vacation-requests-table-toolbar.tsx
```diff
@@ -132,7 +133,19 @@ const TableToolbar = ({
                 setValue={setFormOpen}
               />
             </ToolbarGridItem>
-          )}
+          }
+          {selectedRowIds?.length === 1 && 
+          rows.find((request: VacationsDataGridRow) => request.id === selectedRowIds[0])?.status === VacationRequestStatuses.APPROVED &&
```
Jdallos, line 138: In this case we are only disabling the edit button if the vacation is approved, but I believe we also wanted to disable the edit button if the request was declined. The logic in the previous comment should take care of this

src/components/vacation-requests-table/vacation-requests-table-toolbar/vacation-requests-table-toolbar.tsx
```diff
@@ -123,7 +123,8 @@ const TableToolbar = ({
           >
             <ToolbarDeleteButton setConfirmationHandlerOpen={setConfirmationHandlerOpen} />
           </ToolbarGridItem>
-          {selectedRowIds?.length === 1 && (
+          {selectedRowIds?.length === 1 &&
+          rows.find((request: VacationsDataGridRow) => request.id === selectedRowIds[0])?.status !== VacationRequestStatuses.APPROVED &&
```
Jdallos, line 127: I would say this is a little more inline logic for a conditional render than is ideal, its not especially clear/ readable. I would suggest creating some kind of disableEdit variable which takes the logic e.g.

const disableEdit = selectedRowIds?.length === 1 && rows.find((request: VacationsDataGridRow) => request.id === selectedRowIds[0])?.status !== VacationRequestStatuses.PENDING

You could then pass the disableEdit to the FormToggleButton and you would not need to have a different component for each condition rather a single condition with only the disabled prop changing

Jdallos (changes requested):

src/components/vacation-requests-table/vacation-requests-table-toolbar/vacation-requests-table-toolbar.tsx
```diff
@@ -77,6 +77,7 @@ const TableToolbar = ({
    *
    * @param selectedRowIds selected row ids
    */
+  const disableEdit = selectedRowIds?.length === 1 && rows.find((request: VacationsDataGridRow) => request.id === selectedRowIds[0])?.status !== VacationRequestStatuses.PENDING
```
Jdallos, line 80:
```suggestion
  const disableEdit = selectedRowIds?.length === 1 && rows.find((request: VacationsDataGridRow) => request.id === selectedRowIds[0])?.status !== VacationRequestStatuses.PENDING;
  
```

Jdallos: The suggestion above is adding missing semi colon and a line break before the function.
You have also added  this const between the function (toggleToolbarOpenOnSelectedRowsIds) and its functions comment/ documentation (TSDocs). You should move the disable edit const to be before the function comment

Jdallos: I realise I suggested disableEdit as the name in a previous comment but perhaps isDisabled or disableEditButton is a clearer name than disableEdit.

Jdallos (changes requested):

Jdallos (approved):

Jdallos (approved):
Approved

PR #97, opened 2024-02-22
fixed the loading spinner not present

yuusufisse:


src/components/home/vacations-card.tsx
```diff
@@ -143,6 +143,7 @@ const VacationsCard = () => {
         setError(`${strings.vacationRequestError.fetchRequestError}, ${error}`);
       }
     }
+    setLoading(false);
```
Jdallos, line 146: I agree with updating the loading state here but you can now remove the setLoading(false) from within the try block.

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -52,9 +52,9 @@ const VacationRequestsScreen = () => {
    * Fetch vacation request statuses
    */
   const fetchVacationRequestStatuses = async () => {
+    setLoading(true);
```
Jdallos, line 55: I don't think this is necessary, and typically it's a good idea for loading state to be started from within the try unless there is a reason not to (as in the case of the other places you have made this change).

You should still add the setLoading(false) after the catch as you have done here. Was there no setLoading(false) previously?

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -74,6 +74,7 @@ const VacationRequestsScreen = () => {
       } catch (error) {
         setError(`${strings.vacationRequestError.fetchStatusError}, ${error}`);
       }
+        setLoading(false);
```
Jdallos, line 77: Fix indentation

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -134,10 +135,10 @@ const VacationRequestsScreen = () => {
           });
         }
         setVacationRequests(fetchedVacationRequests);
-        setLoading(false);
       } catch (error) {
         setError(`${strings.vacationRequestError.fetchRequestError}, ${error}`);
       }
+        setLoading(false);
```
Jdallos, line 141: Fix indentation

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -166,11 +167,11 @@ const VacationRequestsScreen = () => {
             (vacationRequestStatus) => vacationRequestStatus.id !== foundVacationRequestStatus.id
           );
           setLatestVacationRequestStatuses(filteredVacationRequestStatuses);
-          setLoading(false);
         }
       } catch (error) {
         setError(`${strings.vacationRequestError.deleteStatusError}, ${error}`);
       }
+        setLoading(false);
```
Jdallos, line 174: Fix indentation

Jdallos (approved):
Approved with comments

Jdallos (approved):
Approved although this may need to be "git merge origin/develop"  as the develop has more recent commits

PR #105, opened 2024-03-11
added lambdas api

DZotoff:


src/api/home-lambdas-api.ts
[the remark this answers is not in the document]
Jdallos: I have found an example here:
https://github.com/Metatavu/etra-user-management/blob/develop/src/api/api.ts

that combines multiple apis into a single file rather that us having the current api.ts and this home-lambdas-api.ts

Seems like a reasonable change to make and shouldn't cause any problems and reduces unnecessary repeated code

src/app/config.ts
```diff
@@ -19,7 +22,8 @@ const env = cleanEnv(import.meta.env, {
   VITE_KEYCLOAK_REALM: str(),
   VITE_KEYCLOAK_CLIENT_ID: str(),
   VITE_API_BASE_URL: url(),
-  VITE_FORECAST_USER_ID_OVERRIDE: num({ default: undefined })
+  VITE_FORECAST_USER_ID_OVERRIDE: num({ default: undefined }),
+  VITE_PIPEDRIVE_URL: url()
```
Jdallos, line 26: This should be named accordingly for the home lambdas url, it is not a pipedrive url which is a seperate service not currently in use here

src/app/config.ts
```diff
@@ -31,6 +35,9 @@ const config: Config = {
   api: {
     baseUrl: env.VITE_API_BASE_URL
   },
+  lambdas: {
+    baseUrl: env.VITE_PIPEDRIVE_URL
```
Jdallos, line 39: Env needs to be renamed

src/atoms/api.ts
```diff
@@ -1,5 +1,7 @@
 import { atom } from "jotai";
 import { getApiClient } from "../api/api";
 import { authAtom } from "./auth";
+import { getLambdasClient } from "../api/home-lambdas-api";
```
Jdallos, line 4: After making changes in the ./auth as described we should be able to import the getLambdasClient from there rather than from here.

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,12 +1,51 @@
...
+        const fetchedTasks = await tasksApi.listProjectTasks({
+          projectId: 418125
+        });
+        
+        const fetchedTimeEntries = await timeEntriesApi.listProjectTimeEntries({
+          projectId: 418125
+        });
+
+      } catch (error) {
+        console.log("error");
```
Jdallos, line 42: This should be console.error

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,12 +1,51 @@
...
+      }
+    }
+  };
```
Jdallos, line 45: It's nice to test this is working but as they are not being used and as they are almost certainly not used in the way they will need to be used I would suggest removing all of the changes in this file.

Jdallos (changes requested):
Requested changes and dependent on other PRs

PR #106, opened 2024-03-21
Feature 100 sprint view screen logic

yuusufisse:


DZotoff (approved):

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,271 @@
...
+        </Card>
+      ) : (
+        <>
+          <FormControlLabel control={<Switch checked={myTasks}/>} label={strings.sprint.showMyTasks}  onClick={() =>  handleOnClick()}/>
+          <FormControl size="small"  style= {{width: "200px", float:"right"}}>
+          <InputLabel disableAnimation={false} >{strings.sprint.taskStatus}</InputLabel>
+            <Select 
+              style={{borderRadius: "30px", marginBottom: "15px" , float:  "right"}}
+              label={strings.sprint.taskStatus}             
+            >
```
Kseniia-Koshkina, line 211: set a default value for select

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,156 @@
...
+ */
+  const getTasksAndTimeEntries = async () => {
+    try {
+      let fetchedTasks = await tasksApi.listProjectTasks({projectId: project.id});
+      if (loggedInpersonId) {
+        fetchedTasks = fetchedTasks.filter((task)=> task.assignedPersons?.includes(loggedInpersonId ));
+      }
+      setTasks(fetchedTasks);
+    /**
+     * Fetch time entries for each task
```
Kseniia-Koshkina, line 52: Probably no need for this commit

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,156 @@
...
+const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
+  const { tasksApi, timeEntriesApi } = useLambdasApi();
+  const [tasks, setTasks] = useState<Tasks[]>([]);
+  const [timeEntries, setTimeEntries] = useState<number[]>([]);
+  const [open, setOpen] = useState<boolean>(false);
+
+/**
+ * Gather tasks and time entries when project is available and update changes appeared
+ */
+  useEffect(() => {
```
Kseniia-Koshkina, line 32:
```suggestion
  /**
   * Gather tasks and time entries when project is available and update changes appeared
   */
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,156 @@
...
+      setTimeEntries(fetchedTimeEntries);
+    } catch (error) {
+      console.error(`Error fetching time entries: ${error}`);
+    }
+  };
+
+/**
+ * Retrieve total time entries for a task
+ */
+  const getTotalTimeEntries = (task: Tasks) => {
```
Kseniia-Koshkina, line 84:
```suggestion
  /**
   * Retrieve total time entries for a task
   */
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,156 @@
...
+        await getTasksAndTimeEntries();
+      }
+    };
+    fetchData();
+  }, [project, loggedInpersonId, filter]);
+
+/**
+ * Get total time entries for the task
+ */
+  const getTasksAndTimeEntries = async () => {
```
Kseniia-Koshkina, line 44:
```suggestion
  /**
   * Get tasks and total time entries
   */
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,156 @@
...
+      }
+      setTasks(fetchedTasks);
+    /**
+     * Fetch time entries for each task
+     */
+      const fetchedTimeEntries = await Promise.all(fetchedTasks.map(async (task) => {
+        try {
+          if (project.id ){           
+            const totalTimeEntries = await timeEntriesApi.listProjectTimeEntries({ projectId: project.id,taskId: task.id });//
+            let totalMinutes = 0;
```
Kseniia-Koshkina, line 58:
```suggestion
            const totalTimeEntries = await timeEntriesApi.listProjectTimeEntries({ projectId: project.id, taskId: task.id });//
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,156 @@
...
+              if (!loggedInpersonId) {
+                totalMinutes += timeEntry.timeRegistered || 0;
+              }
+            })
+            return totalMinutes;
+          }
+          return 0;
+        } catch (error) {
+          console.error(`Error fetching time entries for allocation ${task.id}: ${error}`);
+          return 0;
```
Kseniia-Koshkina, line 72:
```suggestion
          console.error(`Error fetching time entries for task ${task.id}: ${error}`);
```

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,271 @@
...
+          const projectFound = fetchedProjects.find((project)=> project.id === allocation.project);
+          projectFound && projects.push(projectFound);
+        })
+        setProjects(projects);
+      } catch (error) {
+        console.error(error, "Error fetching projects:");
+      }
+    }
+  };
+  const getPersonAllocations = async () => {
```
Kseniia-Koshkina, line 73: missing docs

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,271 @@
...
+        <Card sx={{ p: "25%", display: "flex", justifyContent: "center" }}>
+          <Box sx={{textAlign: 'center'}} >
+          <Typography>{strings.placeHolder.pleaseWait}</Typography>
+          <CircularProgress sx={{ scale: "150%", mt: "5%", mb: "5%"}} />
+          </Box>
+        </Card>
+      ) : (
+        <>
+          <FormControlLabel control={<Switch checked={myTasks}/>} label={strings.sprint.showMyTasks}  onClick={() =>  handleOnClick()}/>
+          <FormControl size="small"  style= {{width: "200px", float:"right"}}>
```
Kseniia-Koshkina, line 206: Probably wrong identations

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,271 @@
...
+                  field: 'allocationsLeft', 
+                  headerName: strings.sprint.allocationLeft, 
+                  flex: 1, cellClassName: (params) =>  timeLeft(params.row) < 0 ? "negative-value" : "", valueGetter: (params) => getHoursAndMinutes(timeLeft(params.row))
+                },
+              ]}             
+            />
+            <Box sx={{ p: 1, display: 'flex',marginLeft:"2px" }}>
+            {strings.sprint.unAllocated}<span style={{paddingLeft: "5px", color: unallocatedTime(allocations) < 0 ? "red" : ""}}> {getHoursAndMinutes(unallocatedTime(allocations))} </span> 
+            </Box>
+          </Card>
```
Kseniia-Koshkina, line 259: identations

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,271 @@
...
+  }
+  
   return (
-    <h1>{ strings.sprint.sprintviewScreen }</h1>
+    <>
+      {loading || !loggedInPerson ? (
+        
+        <Card sx={{ p: "25%", display: "flex", justifyContent: "center" }}>
+          <Box sx={{textAlign: 'center'}} >
+          <Typography>{strings.placeHolder.pleaseWait}</Typography>
```
Kseniia-Koshkina, line 199: identations

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,270 @@
...
+                {strings.sprint.inProgress}
+              </MenuItem>
+              <MenuItem key={3} value={strings.sprint.allTasks} onClick={() => setFilter("DONE")}>
+                {strings.sprint.completed}
+              </MenuItem>
+              <MenuItem key={4} value={strings.sprint.allTasks} onClick={() => setFilter("")}>
+                {strings.sprint.allTasks}
+              </MenuItem>
+            </Select>
+          </FormControl>
```
Jdallos, line 203: This should probably be removed for now as it does not work until there are actually statuses to filer by, which is being done in a future branch I believe

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,270 @@
...
+                  headerName: strings.sprint.timeEntries, 
+                  flex: 1, valueGetter: (params) => getHoursAndMinutes(getTotalTimeEntries(params.row) || 0),
+                },
+                { 
+                  field: 'allocationsLeft', 
+                  headerClassName: 'header-color', 
+                  headerName: strings.sprint.allocationLeft, 
+                  flex: 1, cellClassName: (params) =>  timeLeft(params.row) < 0 ? "negative-value" : "", valueGetter: (params) => getHoursAndMinutes(timeLeft(params.row))
+                },
+              ]}
```
Jdallos, line 250: Improve render method code readability by creating a separate columns variable to contain this. Similar was done for vacations in the vacation-requests-table/vacation-requests-table-columns

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,270 @@
...
+                },
+                { 
+                  field: 'allocationsLeft', 
+                  headerClassName: 'header-color', 
+                  headerName: strings.sprint.allocationLeft, 
+                  flex: 1, cellClassName: (params) =>  timeLeft(params.row) < 0 ? "negative-value" : "", valueGetter: (params) => getHoursAndMinutes(timeLeft(params.row))
+                },
+              ]}             
+            />
+            <Box sx={{ backgroundColor:"#e6e6e6", display: 'flex', justifyContent: "space-between", padding: "5px", paddingTop:" 10px", paddingBottom:" 10px"}}>
```
Jdallos, line 252: Consistent use of double quotes please. A working linter should catch and fix this in the future.

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,270 @@
...
+              <span style={{paddingLeft: "5px", color: unallocatedTime(allocations) < 0 ? "red" : ""}}>{strings.sprint.unAllocated} {getHoursAndMinutes(unallocatedTime(allocations))} </span>
```
Jdallos, line 253:
```suggestion
              <span style={{paddingLeft: "5px", color: unallocatedTime(allocations) < 0 ? "red" : ""}}>{strings.sprint.unAllocated} {getHoursAndMinutes(unallocatedTime(allocations))}</span> 
```

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,270 @@
...
+                  headerName: strings.sprint.allocationLeft, 
+                  flex: 1, cellClassName: (params) =>  timeLeft(params.row) < 0 ? "negative-value" : "", valueGetter: (params) => getHoursAndMinutes(timeLeft(params.row))
+                },
+              ]}             
+            />
+            <Box sx={{ backgroundColor:"#e6e6e6", display: 'flex', justifyContent: "space-between", padding: "5px", paddingTop:" 10px", paddingBottom:" 10px"}}>
+              <span style={{paddingLeft: "5px", color: unallocatedTime(allocations) < 0 ? "red" : ""}}>{strings.sprint.unAllocated} {getHoursAndMinutes(unallocatedTime(allocations))} </span> 
+              <span style={{ paddingRight:"5px"}}>
+                {strings.sprint.currentSprint}: {sprintStartDate.toLocaleString()} - {sprintEndDate.toLocaleString()}
+              </span>
```
Jdallos, line 256: This could all be done with a formatted string and passing the variables to the Strings library rather than having to combine localized strings and colons and dashes as is being done now.
Here is another reference of something similar being done:
https://github.com/Metatavu/oss-surveys-management/blob/53a92e960114a0a5752f282fedc155db252785f7/src/components/screens/overview-screen.tsx#L161

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,11 +1,270 @@
...
+            </Box>
+          </Card>
+          {projects.map((project) => {
+            return (
+              <TaskTable key={project.id} project={project} loggedInpersonId={myTasks ? loggedInPerson?.id : undefined} filter={filter} />        
+            )
+          }
+          )}
```
Jdallos, line 264: Use implicit return here

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+ * Component properties
+ */
+interface Props {
+  project : Projects,
+  loggedInpersonId?: number,
+  filter?: string
+}
+
+/**
+ * Task component
```
Jdallos, line 21: Missing @param component properties DOcs

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
+import { Box, Card, CircularProgress, IconButton } from "@mui/material";
+import { Projects, Tasks, TimeEntries } from "../../generated/homeLambdasClient";
+import { DataGrid } from "@mui/x-data-grid";
+import { useEffect, useState } from "react";
+import { useLambdasApi } from "../../hooks/use-api";
+import { getHoursAndMinutes } from "../../utils/time-utils";
+import KeyboardArrowUpIcon from '@mui/icons-material/KeyboardArrowUp';
```
Jdallos, line 7: Again consistent use of double quotes

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Task component
+ */
+const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
+  const { tasksApi, timeEntriesApi } = useLambdasApi();
+  const [tasks, setTasks] = useState<Tasks[]>([]);
+  const [timeEntries, setTimeEntries] = useState<number[]>([]);
+  const [open, setOpen] = useState<boolean>(false);
+  const [loading, setLoading] = useState<boolean>(false);
```
Jdallos, line 28: Don't need the explicit typing here

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * Gather tasks and time entries when project is available and update changes appeared
+   */
+  useEffect(() => {
+    const fetchData = async () => {
```
Jdallos, line 34: As mentioned in the previous screen, move the loading states into the getTasksAndTimeEntries function and then no need to await in the useEffect

Jdallos: useEffect(() => {
    if (project && open) {
      getTasksAndTimeEntries();
    }
  }, [project, loggedInpersonId, filter, open]);

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+              }
+              if (!loggedInpersonId) {
+                totalMinutes += timeEntry.timeRegistered || 0;
+              }
+            })
+            return totalMinutes;
+          }
+          return 0;
+        } catch (error) {
+          console.error(`Error fetching time entries for task ${task.id}: ${error}`);
```
Jdallos, line 72: Use error handler and localize the error

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+          return 0;
+        }
+      }));
+      setTimeEntries(fetchedTimeEntries);
+    } catch (error) {
+      console.error(`Error fetching time entries: ${error}`);
```
Jdallos, line 78: Use error handler and localilze the error

Jdallos, line 78: Use error handler and localilze the error

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+    }
+  };
+
+  /**
+   * Retrieve total time entries for a task
```
Jdallos, line 83: Missing @param Docs

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+   */
+  const getTotalTimeEntries = (task: Tasks) => {
+    if (timeEntries.length!==0) {
```
Jdallos, line 86:
```suggestion
    if (timeEntries.length) {
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+      return timeEntries[tasks.indexOf(task)];
+    }
+    return 0;
+  }
+
+  /**
+   * Lowercase all letters in a string except the first letter except first one
+   */
+
```
Jdallos, line 95:
```suggestion
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+   */
+  const getTotalTimeEntries = (task: Tasks) => {
+    if (timeEntries.length!==0) {
+      return timeEntries[tasks.indexOf(task)];
+    }
+    return 0;
+  }
+
+  /**
+   * Lowercase all letters in a string except the first letter except first one
```
Jdallos, line 93: Missing @param docs

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+   */
+
+  const lowerCaseAllWordsExceptFirstLetters = (string: string) => {
```
Jdallos, line 96: Perhaps this function could go in utils/localization-utils?

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+    <Card key={0} sx={{ backgroundColor: '#f2f2f2', margin: 0, paddingTop: "5px", paddingBottom: "5px", width: "100%", height: "100", marginBottom: "16px", 
+    '& .high_priority': {
+      color: 'red'}, 
+    '& .low_priority': {
+      color: 'green'} 
+    }}>
+      <IconButton onClick={() => setOpen(!open)}>
+        {open ? <KeyboardArrowUpIcon /> : <KeyboardArrowDownIcon />}
+      </IconButton>
+      <span>{project ? project.name : ""}</span>
```
Jdallos, line 112: Use a Mui Typography component here?

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+                  headerClassName: 'header-color', 
+                  headerName: strings.sprint.estimatedTime, 
+                  flex: 1, valueGetter: (params) => getHoursAndMinutes(params.row.estimate || 0) 
+                },
+                { 
+                  field: 'timeEntries',headerClassName: 'header-color', 
+                  headerName: strings.sprint.timeEntries, 
+                  flex: 1, valueGetter: (params) => getHoursAndMinutes(getTotalTimeEntries(params.row)) 
+                }
+              ]}
```
Jdallos, line 181: Again you could have the columns as a seperate variable to keep the render method as "clean" as possible

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+                },
+                { 
+                  field: 'assignedPersons',
+                  headerClassName: 'header-color',         
+                  headerName: strings.sprint.assigned, 
+                  flex: 1
+                },
+                { 
+                  field: 'status',
+                  headerClassName: 'header-color',
```
Jdallos, line 160: Check for single quotes again, should be using double quotes

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+                  field: 'assignedPersons',
+                  headerClassName: 'header-color',         
+                  headerName: strings.sprint.assigned, 
+                  flex: 1
+                },
+                { 
+                  field: 'status',
+                  headerClassName: 'header-color', 
+                  headerName: strings.sprint.taskStatus, 
+                  flex: 1, valueGetter: (params) => lowerCaseAllWordsExceptFirstLetters(params.row.statusCategory || ''),
```
Jdallos, line 162: If the statusCategory does not exist on the task then this should be removed for now.

src/utils/time-utils.ts
```diff
@@ -171,4 +171,25 @@ const countWorkingWeekDaysInRange = (startWeekIndex: number, endWeekIndex: numbe
...
+  return DateTime.fromISO(date).minus({days : days - 1});
+}
+
+/**
+ * Get sprint end date
+ *
+ * @param date string date 
+ */
+export const getSprintEnd = (date: string) => {
+  return getSprintStart(date).plus({days : 14});
```
Jdallos, line 194: We might want to change this e.g. currently the endDate is the same Monday as the start date of the next sprint, perhaps it should be the Friday before e.g. days: 11?
What do you think is best?

Jdallos (changes requested):
Changes requested

src/components/screens/sprint-view-screen.tsx
DZotoff: It does if update spec

src/components/sprint-view-table/tasks-table.tsx
DZotoff: This function will be removed

src/components/sprint-view-table/tasks-table.tsx
DZotoff: "status" will be used instead of "statusCategory"

src/components/sprint-view-table/tasks-table.tsx
DZotoff: Will be removed, there is no need to have this function and docs

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,20 +1,18 @@
...
+import { useLambdasApi } from "src/hooks/use-api";
+import { Person } from "src/generated/client";
+import { useAtomValue, useSetAtom } from "jotai";
+import { personsAtom,  } from "src/atoms/person";
+import config from "src/app/config";
+import { userProfileAtom } from "src/atoms/auth";
+import { Allocations, Projects, TimeEntries } from "src/generated/homeLambdasClient/models/";
 import { DataGrid } from '@mui/x-data-grid';
-import { getHoursAndMinutes, getSprintEnd, getSprintStart } from '../../utils/time-utils';
+import { getHoursAndMinutes, getSprintEnd, getSprintStart } from 'src/utils/time-utils';
```
Jdallos, line 11: Use double quotes

src/components/screens/sprint-view-screen.tsx
```diff
@@ -29,30 +27,29 @@ const SprintViewScreen = () => {
...
-  const [loading, setLoading] = useState<boolean>(true);
-  const [myTasks, setMyTasks] = useState<boolean>(true);
-  const [filter, setFilter] = useState<string>("");
-  const sprintStartDate = getSprintStart((new Date()).toISOString());
+  const [loading, setLoading] = useState(false);
+  const [myTasks, setMyTasks] = useState(true);
+  const [filter, setFilter] = useState("");
+  const todaysDate = (new Date()).toISOString()
+  const sprintStartDate = getSprintStart(todaysDate);
   const sprintEndDate = getSprintEnd((new Date()).toISOString());
```
Jdallos, line 35: Can't this also use todaysDate?

src/components/screens/sprint-view-screen.tsx
```diff
@@ -29,30 +27,29 @@ const SprintViewScreen = () => {
...
+  const columns = sprintViewProjectsColumns({allocations, timeEntries, projects});
+  const setError = useSetAtom(errorAtom);
 
   /**
-   * Get project data if user is logged in otherwise endless loading
+   * Get project data if user is logged in
    */
   useEffect(() => {
     setLoading(true);
```
Jdallos, line 43: Its best to handle the loading states in one place/ function rather than splitting them between the use effect and the end of the function called in the use effect. In this case I would recommend placing both setLoadings within the function

src/components/screens/sprint-view-screen.tsx
```diff
@@ -29,30 +27,29 @@ const SprintViewScreen = () => {
...
-      fetchViewData();
+      fetchPersonEngagement();
     }
   }, [loggedInPerson]);
 
   /**
    * Fetch allocations, project names and time entries
    */
-  const fetchData = async () => {
+  const fetchPersonEngagement = async () => {
```
Jdallos, line 52: Engagement is a bit vague and doesn't directly relate to the data being fetched.
You could call it something like fetchPersonsProjectsAndAllocationsAndTimeEntries however that's a bit of a mouthful- which is itself an indicator that perhaps this function is doing to much. 
How about splitting it into 3 different functions for fetching projects, allocations and timeEntries. Then you could have a "parent" function which called the three functions in turn (ensuring allocations and projects are available before calling time entries) and handled the loading state (rather than have the loading set for all 3 functions).

Although the current function works I think making the change I suggest improves readability and will make debugging easier in the future.

src/components/screens/sprint-view-screen.tsx
```diff
@@ -71,7 +68,8 @@ const SprintViewScreen = () => {
           }); 
           return totalMinutes;
         } catch (error) {
```
Jdallos, line 70: I can't add a comment on the correct line but this is referring to the above timeEntriesApi.listProjectTimeEntries request.

rather than have projectId: allocation.project || 0
You should check if allocation.project is defined on the first line of the map function before the try, and throw the appropriate error if not. This means you will be able to just do projectId: allocation.project. 
Using the || 0 is unclear.

src/components/screens/sprint-view-screen.tsx
```diff
@@ -71,7 +68,8 @@ const SprintViewScreen = () => {
...
-          console.error(`Error fetching time entries for allocation ${allocation.id}: ${error}`);
+          const message: string = strings.formatString(strings.sprintRequestError.fetchAllocationError, (allocation.id||0).toString(), error as string).toString();
```
Jdallos, line 71: Again the use of allocation.id || 0 even in the error message does not seem right. What does 0 mean, is there a way to say the allocation does not have an id rather than just 0?

src/components/screens/sprint-view-screen.tsx
```diff
@@ -71,7 +68,8 @@ const SprintViewScreen = () => {
...
+          setError(message);
           return 0;
```
Jdallos, line 73: Why are you returning 0 here, if something has caused an error do you have to return anything? If you end up with undefined or something in totalTimeEntries then you could filter out unwanted values before setting it as state.

src/components/screens/sprint-view-screen.tsx
```diff
@@ -86,12 +84,15 @@ const SprintViewScreen = () => {
...
+    setLoading(false);
   };
 
   /**
-   * Calculate project total time spent on the project by the user
+   * Calculate total time allocated to the project for 2 week period
+   * 
+   * @param allocation expected work load of user in minutes
    */
   const totalAllocations = (allocation: Allocations) => {
```
Jdallos, line 97: This is also repeated in sprint-project-columns and should be exported from a utils file to be used in both places

src/components/screens/sprint-view-screen.tsx
```diff
@@ -104,44 +105,9 @@ const SprintViewScreen = () => {
...
-  /**
-   * Get project color
-   */
-  const getProjectColor = (allocation: Allocations) => {
-    if (projects.length !== 0) {
-      return projects[allocations.indexOf(allocation)]?.color;
-    }
-    return "";
-  }
-
```
Jdallos: You moved these in to sprint-project-columns but perhaps they would be better place in a sprint utils file?

src/components/screens/sprint-view-screen.tsx
```diff
@@ -172,19 +138,19 @@ const SprintViewScreen = () => {
...
-          <Box sx={{textAlign: 'center'}} >
+          <Box sx={{ textAlign: "center" }} >
             <Typography>{strings.placeHolder.pleaseWait}</Typography>
-            <CircularProgress sx={{ scale: "150%", mt: "5%", mb: "5%"}} />
+            <CircularProgress sx={{ scale: "150%", mt: "5%", mb: "5%" }} />
           </Box>
         </Card>
       ) : (
         <>
           <FormControlLabel control={<Switch checked={myTasks}/>} label={strings.sprint.showMyTasks}  onClick={() =>  handleOnClickTask()}/>
```
Jdallos, line 148: Loads of components within this render could have their props indented on a new line. Remember as a general rule if a component has more 3 or more props each should be indented on a new line- this massively improves code clarity

src/components/screens/sprint-view-screen.tsx
```diff
@@ -201,66 +167,36 @@ const SprintViewScreen = () => {
...
-                  headerName: strings.sprint.allocationLeft, 
-                  flex: 1, cellClassName: (params) =>  timeLeft(params.row) < 0 ? "negative-value" : "", valueGetter: (params) => getHoursAndMinutes(timeLeft(params.row))
-                },
-              ]}             
+              columns={columns}
             />
-            <Box sx={{ backgroundColor:"#e6e6e6", display: 'flex', justifyContent: "space-between", padding: "5px", paddingTop:" 10px", paddingBottom:" 10px"}}>
-              <span style={{paddingLeft: "5px", color: unallocatedTime(allocations) < 0 ? "red" : ""}}>{strings.sprint.unAllocated} {getHoursAndMinutes(unallocatedTime(allocations))} </span> 
+            <Box sx={{ backgroundColor:"#e6e6e6", display: 'flex', justifyContent: "space-between", padding: "5px", paddingTop:" 10px", paddingBottom:" 10px" }}>
+              <span style={{paddingLeft: "5px", color: unallocatedTime(allocations) < 0 ? "red" : ""}}>{strings.sprint.unAllocated} {getHoursAndMinutes(unallocatedTime(allocations))}</span>
```
Jdallos, line 192: Indent

src/components/sprint-view-table/sprint-tasks-columns.ts
```diff
@@ -0,0 +1,71 @@
...
+      field: "priority",
+      headerClassName: "header-color", 
+      headerName: strings.sprint.taskPriority,
+      cellClassName: (params) => params.row.highPriority ? "high_priority": "low_priority",
+      flex: 1, valueGetter: (params) => params.row.highPriority ? "High" : "Normal"   
+    },
+    { field: "estimate",
+      headerClassName: "header-color", 
+      headerName: strings.sprint.estimatedTime, 
+      flex: 1, valueGetter: (params) => getHoursAndMinutes(params.row.estimate || 0)
```
Jdallos, line 48: Fix indentation

src/components/sprint-view-table/sprint-tasks-columns.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    { field: "estimate",
+      headerClassName: "header-color", 
+      headerName: strings.sprint.estimatedTime, 
+      flex: 1, valueGetter: (params) => getHoursAndMinutes(params.row.estimate || 0) 
+    },
+    { 
+      field: "timeEntries", headerClassName: "header-color", 
+      headerName: strings.sprint.timeEntries, 
+      flex: 1, valueGetter: (params) => getHoursAndMinutes(getTotalTimeEntries(params.row, tasks, timeEntries))
```
Jdallos, line 53: Fix indentation

src/components/sprint-view-table/sprint-tasks-columns.ts
```diff
@@ -0,0 +1,71 @@
...
+      flex: 1, valueGetter: (params) => getHoursAndMinutes(getTotalTimeEntries(params.row, tasks, timeEntries)) 
+    },
+  ];
+  return columns;
+};
+
+/**
+ * Retrieve total time entries for a task
+ * 
+ * @param task task of allocated project
```
Jdallos, line 62: Update @params

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -19,32 +21,32 @@ interface Props {
...
-      setLoading(true);
-      if (project && open) {
-        await getTasksAndTimeEntries();
-        setLoading(false);
-      }
-    };
-    fetchData();
-  }, [project, loggedInpersonId, filter, open]);
+    if (project && open) {
+    getTasksAndTimeEntries();
```
Jdallos, line 41: Indent

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -19,32 +21,32 @@ interface Props {
...
+    }
+    }, [project, loggedInpersonId, filter, open]);
```
Jdallos, line 43: Fix indentation

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -19,32 +21,32 @@ interface Props {
...
 
   /**
    * Get tasks and total time entries  
    */
   const getTasksAndTimeEntries = async () => {
+    setLoading(true);
```
Jdallos, line 49: Do you think we could also have a  check in here to see if the tasks and time entries data has already been loaded e.g. if the user opens and closes the project accordion it re fetches the data, but maybe we don't need to?

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -69,51 +71,34 @@ const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
           }
           return 0;
         } catch (error) {
-          console.error(`Error fetching time entries for task ${task.id}: ${error}`);
+          const message: string = strings.formatString(strings.sprintRequestError.fetchTasksError, task.id||0, error as string).toString();
```
Jdallos, line 74: Can we do something better than task.id ||0 ?

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -69,51 +71,34 @@ const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
...
-  };
-
-  /**
-   * Retrieve total time entries for a task
-   */
-  const getTotalTimeEntries = (task: Tasks) => {
-    if (timeEntries.length!==0) {
-      return timeEntries[tasks.indexOf(task)];
+      setError(`${strings.sprintRequestError.fetchTimeEntriesError} ${error}`);
+      setLoading(false);
```
Jdallos, line 83: No need to repeat the setLoading(false) just have it after the catch rather than in both the try and the catch

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -69,51 +71,34 @@ const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
...
-    });
   };
 
   return ( 
-    <Card key={0} sx={{ backgroundColor: '#f2f2f2', margin: 0, paddingTop: "5px", paddingBottom: "5px", width: "100%", height: "100", marginBottom: "16px", 
-    '& .high_priority': {
-      color: 'red'}, 
-    '& .low_priority': {
-      color: 'green'} 
+    <Card key={0} sx={{ backgroundColor: "#f2f2f2", margin: 0, paddingTop: "5px", paddingBottom: "5px", width: "100%", height: "100", marginBottom: "16px",
```
Jdallos, line 88: Again for all these components fix the indentation, if there are 3 or more props indent each on its own new line

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -69,51 +71,34 @@ const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
...
+    "& .high_priority": {
+      color: "red"}, 
+    "& .low_priority": {
+      color: "green"} 
     }}>
       <IconButton onClick={() => setOpen(!open)}>
         {open ? <KeyboardArrowUpIcon /> : <KeyboardArrowDownIcon />}
       </IconButton>
-      <span>{project ? project.name : ""}</span>
+      <Typography style={{display:"inline"}}>{project ? project.name : ""}</Typography>
```
Jdallos, line 97:
```suggestion
      <Typography style={{display:"inline"}}>{project?.name}</Typography>
```

Jdallos (changes requested):

src/components/screens/sprint-view-screen.tsx
DZotoff: I decided to combine these three functions into one because in opposite way we were increasing time-entries' requests to backend. What's about changing function name on "fetchPersonData" or "fetchProjectDetails"? Function itself looks simplified enough.

src/components/screens/sprint-view-screen.tsx
DZotoff: It affects logic of solution. If we are using filtering before, it breaks the indexing. For example, allocation.length will not be equal to timeEntries.length. Here I simplified returns, deleted not necessary ones.

src/components/sprint-view-table/sprint-projects-columns.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * Component properties
+ */
+interface Props {
+  allocations: Allocations[],
+  timeEntries: number[],
+  projects: Projects[]
+}
+
+/**
+ * Requests data for Project's table columns
```
Jdallos, line 18:
```suggestion
 * Sprint view projects table columns component
```

src/components/sprint-view-table/sprint-projects-columns.tsx
```diff
@@ -0,0 +1,75 @@
...
+          style={{ marginRight:"10px" }} 
+          component="span" 
+          sx={{ 
+            bgcolor: getProjectColor(params.row, allocations, projects), 
+            height: 25, 
+            borderRadius: "5px" 
+          }}
+        />
+        {getProjectName(params.row, allocations, projects)}
+      </>
```
Jdallos, line 47: indent

src/components/sprint-view-table/sprint-tasks-columns.ts
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Component properties
+ */
+interface Props {
+  tasks : Tasks[],
+  timeEntries: number[]
+}
+
+/**
+ * Requests data for Task's table columns
```
Jdallos, line 16:
```suggestion
 * Sprint view tasks table columns component
```

Jdallos: Because this component is not requesting data

src/components/sprint-view-table/sprint-tasks-columns.ts
```diff
@@ -0,0 +1,69 @@
...
+      headerClassName: "header-color",         
+      headerName: strings.sprint.assigned, 
+      flex: 1
+    },
+    { 
+      field: "status",
+      headerClassName: "header-color", 
+      headerName: strings.sprint.taskStatus, 
+      flex: 1, 
+      valueGetter: (params) => params.row.status || "",
```
Jdallos, line 43:
```suggestion
      valueGetter: (params) => params.row.status || ""
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+ * Component properties
+ */
+interface Props {
+  project : Projects,
+  loggedInpersonId?: number,
+  filter?: string
+}
+
+/**
+ * Task component
```
Jdallos, line 23:
```suggestion
 * Task table component
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+import sprintViewTasksColumns from "./sprint-tasks-columns";
+import { errorAtom } from "../../atoms/error";
+import { useSetAtom } from "jotai";
+
+/**
+ * Component properties
+ */
+interface Props {
+  project : Projects,
+  loggedInpersonId?: number,
```
Jdallos, line 18: Use camel case loggedInPersonId

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+ */
+const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
+  const { tasksApi, timeEntriesApi } = useLambdasApi();
+  const [tasks, setTasks] = useState<Tasks[]>([]);
+  const [timeEntries, setTimeEntries] = useState<number[]>([]);
+  const [open, setOpen] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const columns = sprintViewTasksColumns({tasks, timeEntries});
+  const setError = useSetAtom(errorAtom);
+  const [reload, setReload] = useState(false);
```
Jdallos, line 35: I'm very suspicious of this, are you sure its needed? It seems to be tied to the loggedInPersonId changing, but when do we expect that to happen? 
It could be worth us discussing this together to figure out if it can be removed/ replaced with something we already have.

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Gather tasks and time entries when project is available and update changes appeared
```
Jdallos, line 38: What does "update changes appeared" mean?

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+      getTasksAndTimeEntries();
+    }
+  }, [project, open, filter, reload]);
+
+  /**
+   * Handle loggenInPersonId change
+   */
+  useEffect(()=>{
+    setReload(true);
+  },[loggedInpersonId]);
```
Jdallos, line 51:
```suggestion
  }, [loggedInpersonId]);
```

Jdallos: Although I hope this UE can be removed

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+  
+  /**
+   * Get tasks and total time entries  
+   */
+  const getTasksAndTimeEntries = async () => {
+    setLoading(true);
+    if (reload || !tasks.length || !timeEntries.length){
```
Jdallos, line 58: Checking by tasks and timeEntries.length seems reasonable, but I'm not sure about the reload

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+      try {
+        let fetchedTasks = await tasksApi.listProjectTasks({projectId: project.id});
```
Jdallos, line 60: Typically rather than have fetchedTasks as a let and then update it after the filter, fetchedTasks would be a const and then when we filter we create another const e.g. loggedInPersonsTasks (might be an appropriate name in this case)

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+        if (loggedInpersonId) {
+          fetchedTasks = fetchedTasks.filter((task)=> task.assignedPersons?.includes(loggedInpersonId ));
```
Jdallos, line 62:
```suggestion
          fetchedTasks = fetchedTasks.filter((task) => task.assignedPersons?.includes(loggedInpersonId));
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+        }
+        setTasks(fetchedTasks);
+        const fetchedTimeEntries = await Promise.all(fetchedTasks.map(async (task) => {
+          try {
+            if (project.id ){           
+              const totalTimeEntries = await timeEntriesApi.listProjectTimeEntries({ projectId: project.id, taskId: task.id });
+              let totalMinutes = 0;
+              totalTimeEntries.forEach((timeEntry: TimeEntries)=> {
+                if (loggedInpersonId && timeEntry.person===loggedInpersonId) {
+                  totalMinutes+=(timeEntry.timeRegistered || 0)
```
Jdallos, line 72:
```suggestion
                  totalMinutes += (timeEntry.timeRegistered || 0)
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+          fetchedTasks = fetchedTasks.filter((task)=> task.assignedPersons?.includes(loggedInpersonId ));
+        }
+        setTasks(fetchedTasks);
+        const fetchedTimeEntries = await Promise.all(fetchedTasks.map(async (task) => {
+          try {
+            if (project.id ){           
+              const totalTimeEntries = await timeEntriesApi.listProjectTimeEntries({ projectId: project.id, taskId: task.id });
+              let totalMinutes = 0;
+              totalTimeEntries.forEach((timeEntry: TimeEntries)=> {
+                if (loggedInpersonId && timeEntry.person===loggedInpersonId) {
```
Jdallos, line 71:
```suggestion
                if (loggedInpersonId && timeEntry.person === loggedInpersonId) {
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+                }
+                if (!loggedInpersonId) {
+                  totalMinutes += timeEntry.timeRegistered || 0;
+                }
+              })
+              return totalMinutes;
+            }
+          } catch (error) {
+            if (task.id) {
+              const message: string = strings.formatString(strings.sprintRequestError.fetchTasksError, task.id||0, error as string).toString();
```
Jdallos, line 82:
```suggestion
              const message: string = strings.formatString(strings.sprintRequestError.fetchTasksError, task.id || 0, error as string).toString();
```

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+      paddingTop: "5px", 
+      paddingBottom: "5px", 
+      width: "100%", 
+      height: "100", 
+      marginBottom: "16px", 
+    "& .high_priority": {
+      color: "red"}, 
+    "& .low_priority": {
+      color: "green"} 
+    }}>
```
Jdallos, line 110: Fix indentation

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+      <IconButton onClick={() => setOpen(!open)}>
+        {open ? <KeyboardArrowUpIcon /> : <KeyboardArrowDownIcon />}
+      </IconButton>
+      <Typography style={{display:"inline"}}>{project?.name}</Typography>
+      {open && 
+        <>
+          {loading ? 
+            <Box sx={{textAlign: "center"}} >
+              <CircularProgress sx={{
```
Jdallos, line 119: sx on a new line and indent appropriately

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -0,0 +1,160 @@
...
+                "& .header-color": {
+                  backgroundColor: "#f2f2f2"
+                }
+              }}
+              autoHeight={true}
+              localeText={{ noResultsOverlayLabel: strings.sprint.notFound }}
+              disableColumnFilter
+              hideFooter={true}
+              filterModel={
+                {
```
Jdallos, line 143:
```suggestion
              filterModel={{
```

src/utils/sprint-utils.ts
```diff
@@ -0,0 +1,103 @@
+import { Person } from "src/generated/client";
+import { Allocations, Projects, Tasks } from "src/generated/homeLambdasClient";
+
+/**
+ * Retrieve total time entries for an Allocation
+ * 
+ * @param allocation task allocated within a project
```
Jdallos, line 7: Allocations are not really related to tasks. Perhaps just allocation Allocation here would be fine to indicate the name is allocation and its type is Allocation?
Similar change can be made for the other param comments

src/utils/sprint-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @param allocations tasks related to the project
+ * @param timeEntries  total time entries requested for each allocation
```
Jdallos, line 9: This makes sense

src/utils/sprint-utils.ts
```diff
@@ -0,0 +1,103 @@
+import { Person } from "src/generated/client";
+import { Allocations, Projects, Tasks } from "src/generated/homeLambdasClient";
+
+/**
+ * Retrieve total time entries for an Allocation
+ * 
+ * @param allocation task allocated within a project
+ * @param allocations tasks related to the project
```
Jdallos, line 8: I'm not sure about this, again allocations are not tasks so not sure what this means. Also can't we just pass in the given allocation (e.g. the first parameter) so why do we also need the list of all allocations?

src/utils/sprint-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Get project name
+ * 
+ * @param allocation task allocated within a project
+ * @param allocations tasks related to the project
+ * @param projects project associated with the given allocation
+ */
+export const getProjectName = (allocation: Allocations, allocations: Allocations[], projects: Projects[]) => {
```
Jdallos, line 39: Similar question here as above regarding allocation and allocations

src/utils/sprint-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Get project color
+ * 
+ * @param allocation task allocated within a project
+ * @param allocations tasks related to the project
+ * @param projects project associated with the given allocation
+ */
+export const getProjectColor = (allocation: Allocations, allocations: Allocations[], projects: Projects[]) => {
```
Jdallos, line 53: And same question here

src/utils/sprint-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Calculate registered time for the user in the current 2 week period
+ * 
+ * @param person user time spent on the project in minutes
+ */
+
+export const calculateWorkingLoad = (person?: Person) => {
+  if (!person) {
+    return 0;
+  }
```
Jdallos, line 95:
```suggestion
  if (!person) return 0;
 
```

Jdallos (changes requested):
Requested changes

src/components/sprint-view-table/tasks-table.tsx
DZotoff: We give parameter loggedInPersonId to task table component: if it is empty then we get all tasks, else we get specific person tasks. We modified the logic so that now if time entries firstly requested for a person they are not re-requested again. We added reload to fix problem when changes in loggedInPerson force re-requesting of tasks and time-entries.

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -46,45 +47,57 @@ const TaskTable = ({project, loggedInpersonId, filter}: Props) => {
...
+                    totalMinutes += timeEntry.timeRegistered || 0;
+                  }
+                  if (!loggedInPersonId) {
+                    totalMinutes += timeEntry.timeRegistered || 0;
+                  }
+                });
+                return totalMinutes;
+              }
+            } catch (error) {
+              if (task.id) {
```
Jdallos, line 87: We still want an error message even if there is no task Id, so perhaps the condtional should be removed but the contents stay the same, Rather than task.id || 0 could something other than 0 be passed e.g. a localized "task id not found"

Jdallos (approved):
Approved with comments

src/components/constants/index.ts
```diff
@@ -12,4 +12,14 @@ export const COLORS = [
...
+export const daysOfWeek = ['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday'];
+
+/**
+ * Task statuses.
+ * */ 
+export const Status = {
+  TODO: "TODO",
+  INPROGRESS: "INPROGRESS",
+  DONE: "DONE",
+  ALL: ""
```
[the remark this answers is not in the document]
DZotoff: These statuses really exist as status categories in the tasks. We are using these statuses for task filtering by status category. There is no endpoint to retrieve all statuses or status categories in forecast API, there are only TODO, INPROGRESS, DONE listed as existing categories in workflow columns. https://github.com/Forecast-it/API/blob/master/sections/workflow_columns.md#workflow-columns

Kseniia-Koshkina (approved):

src/components/layout/navbar.tsx
```diff
@@ -2,17 +2,29 @@ import { MouseEvent, useState } from "react";
...
   const auth = useAtomValue(authAtom);
   const [anchorElUser, setAnchorElUser] = useState<null | HTMLElement>(null);
+  const avatars: UsersAvatars[] = useAtomValue(avatarsAtom);
+  const persons: Person[] = useAtomValue(personsAtom);
+  const userProfile = useAtomValue(userProfileAtom);
+  const loggedInPerson = persons.find(
+    (person: Person) =>
+      person.id === config.person.forecastUserIdOverride || person.keycloakId === userProfile?.id
+  );
+  const loggedInPersonAvatar = avatars.find(avatar => loggedInPerson?.id===avatar.personId)?.imageOriginal||"";
```
Jdallos, line 27:
```suggestion
  const loggedInPersonAvatar = avatars.find(avatar => loggedInPerson?.id === avatar.personId)?.imageOriginal || "";
```

src/components/providers/authentication-provider.tsx
```diff
@@ -84,8 +86,19 @@ const AuthenticationProvider = ({ children }: Props) => {
...
+  const getSlackAvatars = async () => {
+    const fetchedAvatars = await slackAvatarsApi.slackAvatar();
+    setAvatars(fetchedAvatars);
+  };
+
   useEffect(() => {
-    if (auth) getPersonsList();
+    if (auth) {
+      getPersonsList();
+      getSlackAvatars();
```
Jdallos, line 100: Should this getSlackAvatars really be in the authentication provider, its not related imo. Perhaps this should be in the nav bar?

src/components/sprint-view-table/user-avatars-component.tsx
```diff
@@ -0,0 +1,84 @@
...
+import type { UsersAvatars } from "src/generated/homeLambdasClient";
+
+/**
+ * Component properties
+ */
+interface Props {
+  assignedPersons: number[];
+}
+
+const UserAvatars = ({ assignedPersons }: Props) => {
```
Jdallos, line 14: Missing docs

src/components/sprint-view-table/user-avatars-component.tsx
```diff
@@ -0,0 +1,84 @@
...
+      return (
+        <Tooltip key={personId} title={(person && `${person.firstName} ${person.lastName}`) || ""}>
+          <Avatar src={avatar?.imageOriginal || ""} />
+        </Tooltip>
+      );
+    }
+
+    if (index === maxAvatarsInLine && numberOfAssignedPersons - maxAvatarsInLine > 0) {
+      const groupedPersons = assignedPersons.slice(maxAvatarsInLine);
+      let tooltipTitile = "";
```
Jdallos, line 59: Typo, should be tooltipTitle

src/components/sprint-view-table/user-avatars-component.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    if (index < maxAvatarsInLine) {
+      return (
+        <Tooltip key={personId} title={(person && `${person.firstName} ${person.lastName}`) || ""}>
+          <Avatar src={avatar?.imageOriginal || ""} />
+        </Tooltip>
+      );
+    }
+
+    if (index === maxAvatarsInLine && numberOfAssignedPersons - maxAvatarsInLine > 0) {
```
Jdallos, line 57: This is used in multiple places:
numberOfAssignedPersons - maxAvatarsInLine

So why not make a variable for it?

src/components/sprint-view-table/user-avatars-component.tsx
```diff
@@ -0,0 +1,84 @@
...
+      let tooltipTitile = "";
+
+      groupedPersons.forEach((groupedPersonId: number) => {
+        const personFound = persons.find((person: { id: number }) => person.id === groupedPersonId);
+        if (personFound) {
+          tooltipTitile += `${personFound?.firstName} ${personFound?.lastName}, `;
+        }
+      });
+      tooltipTitile = tooltipTitile.slice(0, tooltipTitile.length - 2);
+      if (numberOfAssignedPersons - maxAvatarsInLine === 1) {
```
Jdallos, line 68: Again, I think a good name for numberOfAssignedPersons - maxAvatarsInLine could make clearer what these conditions are checking for

Jdallos: I'm also not clear why we are specifically checking it === 1 here. Please explain.

src/components/sprint-view-table/user-avatars-component.tsx
```diff
@@ -0,0 +1,84 @@
...
+        return (
+          <Tooltip key={personId} title={tooltipTitile}>
+            <Avatar src={avatar?.imageOriginal} />
+          </Tooltip>
+        );
+      }
+      return (
+        <Tooltip key={"Group of hidden avatars"} title={tooltipTitile}>
```
Jdallos, line 76: This is not an appropriate key, and is not necessarily unique?
How about `hidden-avatars-${index}`

Jdallos (changes requested):
Changes requested

Jdallos (approved):

PR #110, opened 2024-04-02
replaced rome with biome as the linter

yuusufisse:


Kseniia-Koshkina (approved):

biome.json
Jdallos, line 1: You need to add a .vscode folder with an extensions.json file containing:
{
  "recommendations": ["biomejs.biome"]
}


This will then actually apply the linter e.g. correct indentation automatically. You will also notice that it highlights many errors in the code. Some should be fixed others e.g. suggestions about UseEffect dependency lists should be ignored for now, in fact perhaps the config could be adjusted so we don't get dependency list related errors.
If there are any errors you are unsure about just let me know.

Jdallos (changes requested):

Jdallos (approved):

PR #120, opened 2024-04-29
Feature 118 hide vacation card in production based on user role

yuusufisse:


src/components/home/sprint-view-card.tsx
```diff
@@ -8,6 +8,11 @@ import UserRoleUtils from "src/utils/user-role-utils";
  */
 const SprintViewCard = () => {
   const adminMode = UserRoleUtils.adminMode();
+  const userIsDeveloper = UserRoleUtils.isDeveloper();
```
Jdallos, line 11: Is this actually needed? Can't you just use the logic as its set up in the admin-screen and the home-screen and that's enough?

src/App.tsx
```diff
@@ -18,12 +18,14 @@ import { Settings } from "luxon";
...
 import RestrictedContentProvider from "./components/providers/restricted-content-provider";
 import SprintViewScreen from "./components/screens/sprint-view-screen";
+import UserRoleUtils from "./utils/user-role-utils";
 
 /**
  * Application component
  */
 const App = () => {
   const language = useAtomValue(languageAtom);
+  const userIsDeveloper = UserRoleUtils.isDeveloper();
```
Jdallos, line 28: Is this needed here , can't you just use the logic as its set up in the admin-screen and the home-screen and that's enough?

src/components/screens/admin-screen.tsx
```diff
@@ -2,22 +2,26 @@ import { Grid } from "@mui/material";
...
 import SprintViewCard from "../home/sprint-view-card";
+import UserRoleUtils from "src/utils/user-role-utils";
 
 /**
  * Admin screen component
  */
-const AdminScreen = () => (
-  <Grid container spacing={2}>
+const AdminScreen = () => {
+  const userIsDeveloper = UserRoleUtils.isDeveloper();
```
Jdallos, line 11: See comment in slack for suggestion to fix duplication error

src/components/screens/home-screen.tsx
```diff
@@ -2,21 +2,23 @@ import { Grid } from "@mui/material";
 import BalanceCard from "../home/balance-card";
 import VacationsCard from "../home/vacations-card";
 import SprintViewCard from "../home/sprint-view-card";
+import UserRoleUtils from "src/utils/user-role-utils";
 
 /**
  * Home screen component
  */
 const HomeScreen = () => {
+  const userIsDeveloper = UserRoleUtils.isDeveloper();
```
Jdallos, line 11: See comment in slack for suggestion to fix duplication error

src/utils/user-role-utils.tsx
```diff
@@ -30,4 +43,13 @@ export default class UserRoleUtils {
 
     return UserRoleUtils.isAdmin() && adminPathname;
   };
+
+  /**
+   * Check if the logged-in user has devveloper role and is in admin route
+   *
+   * @returns boolean, indicates if user is developer
+   */
+  public static readonly developerMode = () => {
```
Jdallos, line 52: You are not currently using this anywhere. If you want to follow the pattern as is done with AdminMode then you should be calling this in the home-screen and admin-screen rather than isDeveloper

Jdallos (changes requested):

src/components/screens/admin-screen.tsx
```diff
@@ -2,22 +2,28 @@ import { Grid } from "@mui/material";
...
+import UserRoleUtils from "src/utils/user-role-utils";
 
 /**
  * Admin screen component
  */
-const AdminScreen = () => (
-  <Grid container spacing={2}>
+const AdminScreen = () => {
+  const developerMode = UserRoleUtils.developerMode();
+  const sprintViewCard = developerMode ? <SprintViewCard />:null
```
Jdallos, line 12:
```suggestion
  const sprintViewCard = developerMode ? <SprintViewCard /> : null;
```

src/components/screens/admin-screen.tsx
```diff
@@ -2,22 +2,28 @@ import { Grid } from "@mui/material";
...
+  const vacationsCard = developerMode ? <VacationsCard />:null
```
Jdallos, line 13:
```suggestion
  const vacationsCard = developerMode ? <VacationsCard /> : null;
```

Jdallos (approved):

PR #129, opened 2024-06-03
disable rules added biome changes

DZotoff:


src/components/charts/timebank-piechart.tsx
```diff
@@ -42,15 +42,14 @@ const TimebankPieChart = ({ personDailyEntry }: Props) => {
...
+              return null;
             }
+            return renderCustomizedLabel(entry);
           }}
           labelLine={false}
-        >     
+        >
           {dailyEntryToChart(personDailyEntry).map((_entry, index) => (
-            <Cell key={`cell-${index}`} fill={COLORS[index]} />
+            <Cell key={`cell-${_entry.name}`} fill={COLORS[index]} />
```
Jdallos, line 52: Are you sure entry.name is unique. Do we have entry.id?
Also _ is used to indicate an unused argument, as it is being used here now we can remove the _

src/components/screens/timebank-view-all-screen.tsx
```diff
@@ -107,7 +117,7 @@ const TimebankViewAllScreen = () => {
             label={(props) => getHoursAndMinutes(props.value)}
           >
             {workTimeData.map((_entry, index) => (
-              <Cell key={`cell-${index}`} fill={COLORS[index % COLORS.length]} />
+              <Cell key={`cell-${_entry.name}`} fill={COLORS[index % COLORS.length]} />
```
Jdallos, line 120: Again no need for _ if we are using entry. Do we not have entry.id?

Jdallos (approved):
Approved with a couple of minor comments

src/components/charts/timebank-piechart.tsx
DZotoff: There is no entry.id property

PR #132, opened 2024-06-04
Metatavu home UI- timebank select entry bug

DZotoff:
Fixed DatePicker component bug and pie chart bug ("Encountered two children"  warning).


Essomatic (approved):

yuusufisse (approved):
looks good to me

src/components/timebank/specific-time-entries-card.tsx
```diff
@@ -63,6 +63,8 @@ const SpecificTimeEntriesCard = ({ selectedEmployeeId }: Props) => {
     end: todayOrEarlier
   });
 
+  const [open, setOpen] = useState(false);
```
Jdallos, line 66: This seems like a fine solution, can you also implement the same solution for the date pickers if the user clicks the by range checkbox

PR #138, opened 2024-06-12
Bugfix 136 sync button error

Essomatic:
sync button crash


src/components/layout/sync-button.tsx
```diff
@@ -20,61 +21,49 @@ const SyncButton = () => {
   const [syncing, setSyncing] = useState(false);
   const [syncSuccess, setSyncSuccess] = useState(false);
   const [syncHandlerOpen, setSyncHandlerOpen] = useState(false);
+  const [canSync, setCanSync] = useState(false);
```
Jdallos, line 24: perhaps enableSync is the more appropriate term?

src/components/layout/sync-button.tsx
```diff
@@ -20,61 +21,49 @@ const SyncButton = () => {
...
       setError(`${strings.error.fetchFailedNoEntriesGeneral}, ${error}`);
     }
+    setCanSync(true);
   };
 
+  useEffect(() => {
+    if(persons.length)
+      {
+        fetchDailyEntries();
+      }
```
Jdallos, line 51: Fix indentation

Jdallos: Do you not have biome enabled?

src/components/layout/sync-button.tsx
```diff
@@ -20,61 +21,49 @@ const SyncButton = () => {
...
-
-      if (fetchedDailyEntries?.length) {
-        dailyEntryDates = fetchedDailyEntries.map((dailyEntry) =>
-          DateTime.fromJSDate(dailyEntry.date)
-        );
+      if (dailyEntries?.length) {
+        dailyEntryDates = dailyEntries.map((dailyEntry) => DateTime.fromJSDate(dailyEntry.date));
       }
+    } catch (error) {
+      console.error("Error fetching daily entries:", error);
```
Jdallos, line 65: We generally don't use console.error in the project as we have the setError which handles errros. So I think this can be removed, if not it should be localized

Jdallos (approved):
Couple of suggestions and approved

Iiro127 (approved):
Approved

PR #153, opened 2024-07-09
Hotfix 152 hotfix isdeveloper

Essomatic:
Instructions: Hotfix to production, use isDeveloper feature (currently only in develop) to disable the balance card (as due to severa change it's now broken). This means we now have nothing in metatavu home production.

### DIFF
--- a/src/utils/user-role-utils.tsx
+++ b/src/utils/user-role-utils.tsx
@@ -5,7 +5,9 @@ import { useLocation } from "react-router";
 /**
  * User Role Utils class
  */
-export default class UserRoleUtils {
+
+// biome-ignore lint/complexity/noStaticOnlyClass: <explanation>
+export  default class UserRoleUtils {
   /**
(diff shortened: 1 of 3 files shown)


src/utils/user-role-utils.tsx
```diff
@@ -30,4 +45,13 @@ export default class UserRoleUtils {
 
     return UserRoleUtils.isAdmin() && adminPathname;
   };
+
+  /**
+   * Check if the logged-in user has devveloper role and is in admin route
```
Kseniia-Koshkina, line 50: double v

Kseniia-Koshkina: Also it would be good to reduce the duplication of code

Essomatic: > Also it would be good to reduce the duplication of code

I fixed the typo, but the code is copied from current development to master, should be probably refactored in develop not here since its just a hotfix.

PR #175, opened 2024-10-28
Create & Save questionnaire

dalju005-xamk:
Function handleSaveSubmit. 
Added Description field, like it is required in swagger. 
Loading for saveSubmit feature.

### DIFF
--- a/src/components/questionnaire/new-questionnaire-card.tsx
+++ b/src/components/questionnaire/new-questionnaire-card.tsx
@@ -12,7 +12,6 @@ import {
 import type React from "react";
 import { useState } from "react";
 import strings from "src/localization/strings";
-import type { QuestionOption } from "src/types/index";
 
 /**
  * Interface for the NewQuestionnaireCard component
@@ -27,7 +26,7 @@ interface Props {
  */
 const NewQuestionnaireCard = ({ handleAddQuestion }: Props) => {
   const [questionText, setQuestionText] = useState("");
-  const [options, setOptions] = useState<QuestionOption[]>([{ label: "", value: false }]);
+  const [options, setOptions] = useState([{ label: "", value: false }]);
 
   /**
    * Handle options label (answer option) change
--- a/src/components/questionnaire/new-questionnaire-card.tsx
+++ b/src/components/questionnaire/new-questionnaire-card.tsx
@@ -22,6 +22,7 @@ interface Props {
 
 /**
  * New Questionnaire Card Component
+ * 
  * @params handleAddQuestion
  */
 const NewQuestionnaireCard = ({ handleAddQuestion }: Props) => {
@@ -30,6 +31,7 @@ const NewQuestionnaireCard = ({ handleAddQuestion }: Props) => {
 
   /**
    * Handle options label (answer option) change
+   * 
    * @param index
    */
   const handleAnswerLabelChange = (
@@ -43,6 +45,7 @@ const NewQuestionnaireCard = ({ handleAddQuestion }: Props) => {
 
   /**
    * Handle options value change (checkbox)
+   * 
    * @param index
    */
   const handleCheckboxChange = (index: number) => {
--- a/src/components/questionnaire/new-questionnaire-card.tsx
+++ b/src/components/questionnaire/new-questionnaire-card.tsx
@@ -17,21 +17,15 @@ import strings from "src/localization/strings";
  * Interface for the NewQuestionnaireCard component
  */
 interface Props {
-  handleAddQuestionSubmit: (
-    questionText: string,
-    options: { label: string; value: boolean }[]
-  ) => void;
+  handleAddQuestionSubmit: (question: string, options: { label: string; value: boolean }[]) => void;
 }
 
 /**
  * New Questionnaire Card Component
  */
-const NewQuestionnaireCard = ({handleAddQuestionSubmit}: Props) => {
-  
-  const [questionText, setQuestionText] = useState("");
-  const [options, setOptions] = useState([
-    { label: "", value: false }
-  ]);
+const NewQuestionnaireCard = ({ handleAddQuestionSubmit }: Props) => {
+  const [question, setQuestionText] = useState("");
+  const [options, setOptions] = useState([{ label: "", value: false }]);
 
   /**
    * Handle options label (answer option) change
@@ -67,12 +61,9 @@ const NewQuestionnaireCard = ({handleAddQuestionSubmit}: Props) => {
    * Handle saving the question (submitting the question and options + resetting the form)
    */
   const handleSaveQuestion = () => {
-    handleAddQuestionSubmit(questionText, options);
-
+    handleAddQuestionSubmit(question, options);
     setQuestionText("");
-    setOptions([
-      { label: "", value: false }
-    ]);
+    setOptions([{ label: "", value: false }]);
   };
 
   return (
@@ -96,15 +87,13 @@ const NewQuestionnaireCard = ({handleAddQuestionSubmit}: Props) => {
             label={strings.newQuestionnaireCard.questionLabel}
             multiline
             rows={6}
-            value={questionText}
+            value={question}
             onChange={(e) => setQuestionText(e.target.value)}
             fullWidth
           />
-
           <Typography variant="body1" sx={{ mt: 2 }}>
             {strings.newQuestionnaireCard.correctAnswer}
           </Typography>
-
           {options.map((option, index) => (
             <Box
               key={index}
@@ -145,7 +134,9 @@ const NewQuestionnaireCard = ({handleAddQuestionSubmit}: Props) => {
             </Box>
           ))}
           <Button onClick={handleAddNewOption} sx={{ mt: 3 }}>
-            <Typography sx={{ fontWeight: "bold", mb: 2 }}>{strings.newQuestionnaireCard.addAnswer}</Typography>
+            <Typography sx={{ fontWeight: "bold", mb: 2 }}>
+              {strings.newQuestionnaireCard.addAnswer}
+            </Typography>
           </Button>
           <CardActionArea>
             <CardActions>
--- a/src/components/questionnaire/new-questionnaire-card.tsx
+++ b/src/components/questionnaire/new-questionnaire-card.tsx
@@ -26,8 +26,8 @@ interface Props {
  * @params handleAddQuestionSubmit
  */
 const NewQuestionnaireCard = ({ handleAddQuestionSubmit }: Props) => {
-  const [question, setQuestionText] = useState("");
-  const [options, setOptions] = useState<QuestionOption[]>([]);
+  const [questionText, setQuestionText] = useState("");
+  const [options, setOptions] = useState<QuestionOption[]>([{label: "", value: false}]);
 
   /**
    * Handle options label (answer option) change
@@ -63,7 +63,7 @@ const NewQuestionnaireCard = ({ handleAddQuestionSubmit }: Props) => {
    * Handle adding new question (submitting the question and options + resetting the form)
    */
   const handleAddNewQuestion = () => {
-    handleAddQuestionSubmit(question, options);
+    handleAddQuestionSubmit(questionText, options);
     setQuestionText("");
     setOptions([{ label: "", value: false }]);
   };
@@ -89,7 +89,7 @@ const NewQuestionnaireCard = ({ handleAddQuestionSubmit }: Props) => {
             label={strings.newQuestionnaireCard.questionLabel}
             multiline
             rows={6}
-            value={question}
+            value={questionText}
             onChange={(e) => setQuestionText(e.target.value)}
             fullWidth
           />
--- a/src/components/questionnaire/new-questionnaire-card.tsx
+++ b/src/components/questionnaire/new-questionnaire-card.tsx
@@ -12,6 +12,7 @@ import {
 import type React from "react";
 import { useState } from "react";
 import strings from "src/localization/strings";
+import type { QuestionOption } from "src/types/index";
 
 /**
  * Interface for the NewQuestionnaireCard component
@@ -22,10 +23,11 @@ interface Props {
 
 /**
  * New Questionnaire Card Component
+ * @params handleAddQuestionSubmit
  */
 const NewQuestionnaireCard = ({ handleAddQuestionSubmit }: Props) => {
   const [question, setQuestionText] = useState("");
-  const [options, setOptions] = useState([{ label: "", value: false }]);
+  const [options, setOptions] = useState<QuestionOption[]>([]);
 
   /**
    * Handle options label (answer option) change
@@ -58,9 +60,9 @@ const NewQuestionnaireCard = ({ handleAddQuestionSubmit }: Props) => {
   };
 
   /**
-   * Handle saving the question (submitting the question and options + resetting the form)
+   * Handle adding new question (submitting the question and options + resetting the form)
    */
-  const handleSaveQuestion = () => {
+  const handleAddNewQuestion = () => {
     handleAddQuestionSubmit(question, options);
     setQuestionText("");
     setOptions([{ label: "", value: false }]);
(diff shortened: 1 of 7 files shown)


naStipe (approved):
Looks good

Hanhz2707 (approved):

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
-  
+  const { questionnaireApi } = useLambdasApi();
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
   const [questionnaireTitle, setQuestionnaireTitle] = useState("");
+  const [questionnaireDescription, setQuestionnaireDescription] = useState("");
+  const [options, setOptions] = useState<Question[]>([]);
+  const [passScoreValue, setPassScoreValue] = useState(0);
 
   const handleQuestionTitleChange = (event: ChangeEvent<HTMLInputElement>) => {
```
Jdallos, line 41: You could probably do something nice here where the name is passed from the textField and used in the  handler to update a single questionnaire state containing title, description and passscore- I think the options could also be updated in this way.

This would mean you have a single questionnaire state and a single handler, which updates the state according to the name attribute from the textfield component

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
+
+  const handlePassScoreSliderChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setPassScoreValue(Number(event.target.value));
+  };
 
   /**
    * Functions to add new question to Questionnaire
-   * @param questionText 
-   * @param options 
+   * @param question
```
Jdallos, line 55:
```suggestion
   * @param question string
```

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
+   * @param options
```
Jdallos, line 56: list of QuestionsOptions

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
-  const handleAddQuestionSubmit = (
-    questionText: string,
-    options: QuestionOption[]
-  ) => {
-    setQuestions((prevQuestions: Question[]) => [...prevQuestions, { questionText, options }]);
+  const handleAddQuestionSubmit = (question: string, options: QuestionOption[]) => {
+    setOptions((prevQuestions: Question[]) => [...prevQuestions, { question, options }]);
   };
 
   const handleDeleteQuestion = (index: number) => {
```
Jdallos, line 62: Missing Docs

Jdallos: Is this deleting questions or options, we need consistant naming

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
+    }, 0);
   };
 
   /**
-   * This should save the question title and questions to DB.
-   * TODO:
+   * Function to save the new questionnaire
    */
-  const handleSaveSubmit = () => {};
+  const handleSaveSubmit = async () => {
```
Jdallos, line 80: Would saveQuestionnaire be a good name for this?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
+    try {
+      setLoading(true);
+      setError(null);
+
+      const createdQuestionnaire = await questionnaireApi.createQuestionnaire({
+        questionnaire: {
+          id: "",
```
Jdallos, line 87: Do we have to pass an empty string here? I would hope the lambda would assign the id, but I would also hope we don't need to provide id as empty string at this point, if we do we might need to update the specs to the type of id is optional?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
+          passScore: passScoreValue
+        }
+      });
+      return createdQuestionnaire;
+    } catch (error) {
+      console.error("Failed to save questionnaire", error);
+      setError("Failed to save questionnaire. Please try again.");
+      setTimeout(() => {
+        setError(null);
+      }, 3000);
```
Jdallos, line 100: Hopefully use of the error handler can handle all of this. Also the Error string should come from the localized strings files same as all other error strings

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,48 +22,90 @@ import { KeyboardReturn } from "@mui/icons-material";
...
+    } finally {
+      setLoading(false);
+      setQuestionnaireTitle("");
+      setQuestionnaireDescription("");
+      setOptions([]);
+      setPassScoreValue(0);
```
Jdallos, line 106: This might require some refactoring if you update the state as I suggest in a previous comment. It might also be worth grouping these into a seperate "closeAndClear" function. Especially if we think such resetting of the state might end up needing to happen in multiple places?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -73,26 +130,65 @@ const NewQuestionnaireScreen = () => {
...
+                sx={{ display: "flex", alignItems: "center", mb: 1, mt: 3 }}
+              >
+                {strings.newQuestionnaireScreen.countedAnswers} {countCorrectAnswers()}
+              </Typography>
+              <Typography variant="h6" gutterBottom sx={{ mb: 1, mt: 1 }}>
+                {strings.newQuestionnaireScreen.requiredAnswers} {passScoreValue}
+              </Typography>
+              <Slider
+                value={passScoreValue}
+                onChange={handlePassScoreSliderChange as any}
```
Jdallos, line 168: Does this have to be cast as any?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -108,7 +204,7 @@ const NewQuestionnaireScreen = () => {
           </Button>
         </Link>
       </Card>
-      {/* Card containing all the build questions */}
+      {/* Card containing all the build options */}
```
Jdallos, line 207: Build options? Don't you mean questionnaire questions?

src/localization/en.json
```diff
@@ -241,7 +241,11 @@
     "preview": "Questionnaire preview",
     "is": "is",
     "delete": "Delete Question",
-    "back": "Back"
+    "back": "Back",
+    "description": "Description",
+    "insertDescription": "Insert Description Here",
```
Jdallos, line 246:
```suggestion
    "insertDescription": "Insert description",
```

src/localization/en.json
```diff
@@ -241,7 +241,11 @@
...
+    "countedAnswers": "True values in this Questionnaire:",
```
Jdallos, line 247: I'm not sure what this string is actually indicating. Do you mean correct answers? Number of correct answers in this questionnnaire? That seems like an off string, surely the number of correct answers/ true values is the same as the number of questions?

src/localization/en.json
```diff
@@ -241,7 +241,11 @@
...
+    "requiredAnswers": "Set Required Correct Answers:"
```
Jdallos, line 248:
```suggestion
    "requiredAnswers": "Set required correct answers:"
```

Jdallos: Don't need to capitalize the start of each word

src/types/index.ts
```diff
@@ -172,7 +172,7 @@ export interface QuestionOption {
  * Type describing a question with text and options in New Quiz Screen
  */
 export interface Question {
-  questionText: string;
+  question: string;
```
Jdallos, line 175: Was there a reason to change from questionText to question. As the interface is called Question it seems reasonable to me to call the attribute questionText

Jdallos (changes requested):

src/components/screens/new-questionnaire-screen.tsx
dalju005-xamk: Well this is to delete already built question before Questionnaire is saved. In UI admin needs to build questions one by one. By saving question it is then rendered in preview. So this is a possibility to delete question from preview before the hole Questionnaire is saved.

Also I now think our naming specs suck... Options in options...

src/components/screens/new-questionnaire-screen.tsx
dalju005-xamk: Bugfix helped

src/components/questionnaire/new-questionnaire-card.tsx
```diff
@@ -17,21 +17,15 @@ import strings from "src/localization/strings";
...
  */
-const NewQuestionnaireCard = ({handleAddQuestionSubmit}: Props) => {
-  
-  const [questionText, setQuestionText] = useState("");
-  const [options, setOptions] = useState([
-    { label: "", value: false }
-  ]);
+const NewQuestionnaireCard = ({ handleAddQuestionSubmit }: Props) => {
+  const [question, setQuestionText] = useState("");
+  const [options, setOptions] = useState([{ label: "", value: false }]);
```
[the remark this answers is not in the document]
dalju005-xamk: Change typing. Would like to keep this default value, so it would have one empty QuestionOption available at first. If there is better way to handle this, let me know.

src/components/screens/new-questionnaire-screen.tsx
dalju005-xamk: I was able to cast both textfield together. But Slider was a problem so I leave it as it was. Also there are no tags, so that will be task for future.

src/components/screens/new-questionnaire-screen.tsx
dalju005-xamk: Now in useState declarations I define what are values when page mounts. Would it be useful to use useEffect() and this closeAndClear() to set those empty values in page loading or is this setup ok ?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -114,16 +112,15 @@ const NewQuestionnaireScreen = () => {
...
-      console.error("Failed to save questionnaire", error);
-      setError("Failed to save questionnaire. Please try again.");
+      setError(`${strings.error.questionnaireSaveFailed}, ${error}`);
       setTimeout(() => {
-        setError(null);
-      }, 3000);
-    } finally {
-      closeAndClear();
+        setError(undefined);
+      }, 5000);
```
Jdallos, line 121: I'm not sure the error handler should be handled in this way. If the error handler is triggered we want the user  to close the dialog manually, rather than it close after the timeout

src/localization/en.json
```diff
@@ -29,7 +29,8 @@
     "dailyEntriesNotFound": "Could not find daily entries",
     "fetchFailedGeneral": "There was an error fetching data",
     "fetchFailedNoEntriesGeneral": "No time entries found",
-    "fetchSlackAvatarsFailed": "Error while fetching Slack avatars"
+    "fetchSlackAvatarsFailed": "Error while fetching Slack avatars",
+    "questionnaireSaveFailed": "Saving questionnaire has failed"
```
Jdallos, line 33: Sticking with the usual structure of an error message I'd suggest:
"Error while saving questionnaire"

Jdallos (approved):

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -7,52 +21,111 @@ import { KeyboardReturn } from "@mui/icons-material";
...
+   * @param event
+   */
+  const handleQuestionnaireInputChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = event.target;
+    if (name === "title") {
+      setQuestionnaireTitle(value);
+    } else if (name === "description") {
+      setQuestionnaireDescription(value);
+    }
   };
```
Jdallos, line 52: I'll expand on my previous suggestion as there is a cleaner way of doing this:

First you should have a single questionnaireState something like this:
const [questionnaire, setQuestionnaire] = useState<Questionnaire>({
    description: "",
    options: [],
    title: "",
    passScore: 0
  });
  
  Then the handler function would be like this:
   /**
   * Function to handle input change in the questionnaire title and description
   * @param event
   */
  const handleQuestionnaireInputChange = (event: ChangeEvent<HTMLInputElement>) => {
    const { name, value } = event.target;

    setQuestionnaire((prevQuestionnaire) => ({ ...prevQuestionnaire, [name]: value }));
  };
  
  Try implementing and testing this. If anything is unclear just ask

Jdallos (changes requested):

src/components/screens/new-questionnaire-screen.tsx
Jdallos: I have added another comment to give an example of how I would do this. Hopefully you can resolve the Slider issue (I guess by slider you are referring to the pass score value state?)

src/components/screens/new-questionnaire-screen.tsx
Jdallos: You might need to tweak this a little bit to handle options but should still be quite simple

src/components/screens/new-questionnaire-screen.tsx
Jdallos: From what I understand we are removing the option from state here, but not actually deleting anything that has been persisted to the DB. Might be worth trying to clarify that in the naming if possible. Maybe its not necessary- your call

src/components/screens/new-questionnaire-screen.tsx
dalju005-xamk: Jeah, it might be better. Now dialog needs to be closed manually after unsuccessful saving has accrued

src/components/screens/new-questionnaire-screen.tsx
dalju005-xamk: I think state management is working now, but problem accured when saving to DB. It worked before. Try to fix this, but might need checking it together.

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -66,24 +70,31 @@ const NewQuestionnaireScreen = () => {
    * @param list of QuestionOptions
    */
   const handleAddQuestion = (questionText: string, options: QuestionOption[]) => {
-    setOptions((prevQuestions: Question[]) => [...prevQuestions, { questionText, options }]);
+    setQuestionnaire((prevQuestionnaire) => ({
+      ...prevQuestionnaire,
+      options: [
```
Jdallos, line 75: Maybe it's a bit strange options are called options, as to me it implies its just the options for a question. Should it be called questions, as it includes the question text and the options? What do you think?

Jdallos: I realize this would also require a change to the specs and the lambdas, so maybe it's not necessary. Perhaps if you find yourself with spare time as I think the naming would be clearer, especially as the Type of options in Questionnaire is Questions (but it's called Questions), and within the questions type we have the Options type which is a label and a value.....

So actually this should be changed as the naming is quite confusing

src/types/index.ts
```diff
@@ -176,3 +176,12 @@ export interface Question {
   options: QuestionOption[];
 }
 
+/**
+ * Type describing a questionnaire in New Questionnaire Screen
+ */
+export interface Questionnaire {
```
Jdallos, line 182: We should be able to get rid of the Questionnaire and Question type in this types file as we can use the types from the generated files. This also gives us better type safety and should make it easier to notice any mis-matches between the spec and the UI

Jdallos (approved):

src/components/screens/new-questionnaire-screen.tsx
dalju005-xamk: Naming is confusing, I agree with that. Perhaps this could be done as seperate fix PR.

PR #178, opened 2024-11-04
Feature 176 questionnaire UI List&Delete

dalju005-xamk:
Listing questionnaires to x-grid-table and added possibility to delete questionnaire.

### DIFF
--- a/src/components/questionnaire/questionnaire-table.tsx
+++ b/src/components/questionnaire/questionnaire-table.tsx
@@ -27,7 +27,7 @@ import { errorAtom } from "src/atoms/error";
  */
 const QuestionnaireTable = () => {
   const adminMode = UserRoleUtils.adminMode();
-  const { questionnaireApi } = useLambdasApi();
+  const { questionnairesApi } = useLambdasApi();
   const [questionnaires, setQuestionnaires] = useState<Questionnaire[]>([]);
   const [loading, setLoading] = useState(false);
   const [dialogOpen, setDialogOpen] = useState(false);
@@ -39,7 +39,7 @@ const QuestionnaireTable = () => {
     const fetchQuestionnaires = async () => {
       setLoading(true);
       try {
-        const questionnaires = await questionnaireApi.listQuestionnaires();
+        const questionnaires = await questionnairesApi.listQuestionnaires();
         setQuestionnaires(questionnaires);
       } catch (error) {
         setError(`${strings.error.questionnaireLoadFailed}, ${error}`);
@@ -77,7 +77,7 @@ const QuestionnaireTable = () => {
       return;
     setLoading(true);
     try {
-      await questionnaireApi.deleteQuestionnaire({ id: deleteId });
+      await questionnairesApi.deleteQuestionnaires({ id: deleteId });
       setQuestionnaires((prevQuestionnaires: Questionnaire[]) =>
         prevQuestionnaires.filter((questionnaire: Questionnaire) => questionnaire.id !== deleteId)
       );
--- a/src/components/questionnaire/questionnaire-table.tsx
+++ b/src/components/questionnaire/questionnaire-table.tsx
@@ -10,8 +10,6 @@ import {
 } from "@mui/material";
 import EditIcon from "@mui/icons-material/Edit";
 import DeleteForeverIcon from "@mui/icons-material/DeleteForever";
-import CheckCircleIcon from "@mui/icons-material/CheckCircle";
-import CloseIcon from "@mui/icons-material/Close";
 import { useState, useEffect } from "react";
 import type { Questionnaire } from "src/generated/homeLambdasClient";
 import strings from "src/localization/strings";
@@ -23,6 +21,7 @@ import { errorAtom } from "src/atoms/error";
 
 /**
  * Questionnaire Table Component
+ *
  * @returns Questionnaires from DynamoDB rendered in a x-data-grid table
  */
 const QuestionnaireTable = () => {
@@ -51,6 +50,7 @@ const QuestionnaireTable = () => {
 
   /**
    * Function to open the dialog for deleting the questionnaire
+   *
    * @param id - Questionnaire ID
    * @param title - Questionnaire Title
    */
@@ -60,9 +60,9 @@ const QuestionnaireTable = () => {
     setDialogOpen(true);
   };
 
-/**
- * Function to close the dialog for deleting the questionnaire
- */
+  /**
+   * Function to close the dialog for deleting the questionnaire
+   */
   const handleCloseDialog = () => {
     setDialogOpen(false);
     setDeleteId(null);
@@ -73,8 +73,7 @@ const QuestionnaireTable = () => {
    * Function to delete the questionnaire
    */
   const handleConfirmDelete = async () => {
-    if (!deleteId) 
-      return;
+    if (!deleteId) return;
     setLoading(true);
     try {
       await questionnairesApi.deleteQuestionnaires({ id: deleteId });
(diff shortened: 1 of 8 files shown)


Hanhz2707 (approved):

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+              </Button>
+            </>
+          )
+        }
+      : {
+          field: "status",
+          headerName: `${strings.questionnaireTable.status}`,
+          flex: 1,
+          renderCell: (params: GridRenderCellParams) =>
+            params.row.passedUsers && params.row.passedUsers.lenght > 0 ? (
```
Jdallos, line 106: Am I correct in thinking we want the check if the logged in user is a passed user? Isn't this rendering the check if there is any passedUser? If so this should be checking if loggedInUser.id is in passedUsers

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+              left: "50%",
+              transform: "translate(-50%, -50%)"
+            }}
+          />
+        )}
+        <DataGrid
+          rows={questionnaires}
+          columns={columns}
+          pagination
+          pageSizeOptions={[5]}
```
Jdallos, line 132: Would be nice to let the user customize this- should be built in option with the data Grid already

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+        </DialogContent>
+        <DialogActions>
+          <Button onClick={handleCloseDialog} color="primary">
+            {strings.questionnaireTable.cancel}
+          </Button>
+          <Button onClick={handleConfirmDelete} color="secondary" variant="contained">
+            {strings.questionnaireTable.confirm}
+          </Button>
+        </DialogActions>
+      </Dialog>
```
Jdallos, line 150: Might be worth having this as the return value of a renderConfirmDeleteDialog, will clean up the main component return. It could also be completly exported into it's own component, but if we are only using this in this one place then perhaps this is not necessary.

src/components/screens/questionnaire-screen.tsx
```diff
@@ -1,40 +1,18 @@
...
-import DeleteForeverIcon from "@mui/icons-material/DeleteForever";
-import EditIcon from "@mui/icons-material/Edit";
+import { Card, CircularProgress, Grid } from "@mui/material";
 import { Typography, Button } from "@mui/material";
 import { Link } from "react-router-dom";
 import UserRoleUtils from "src/utils/user-role-utils";
 import { KeyboardReturn } from "@mui/icons-material";
 import strings from "src/localization/strings";
+import QuestionnaireTable from "../questionnaire/questionnaire-table";
+import { Suspense } from "react";
```
Jdallos, line 8: Haven't seen this used much before- is it really necessary here?

src/components/screens/questionnaire-screen.tsx
```diff
@@ -81,35 +58,12 @@ const QuestionnaireScreen = () => {
...
-                  </Button>
-                </CardActions>
-              }
-            </Card>
-          </Grid>
-        ))}
       </Grid>
+      <Suspense fallback={<CircularProgress/>}>
+        <QuestionnaireTable/>
+      </Suspense>
```
Jdallos, line 64: Couldn't there just be a renderLoader as in other components? Or do you think suspense is better? Seems to work from in the UI.

src/components/screens/questionnaire-screen.tsx
```diff
@@ -81,35 +58,12 @@ const QuestionnaireScreen = () => {
...
+      
+
```
Jdallos, line 66:
```suggestion
```

Jdallos: No need for double line breaks

src/localization/en.json
```diff
@@ -250,5 +250,17 @@
...
     "saveAnswer": "Save question"
+  },
+  "questionnaireTable": {
+    "title": "Questionnaire Title",
+    "description": "Description",
+    "actions": "Actions",
+    "edit": "Edit",
+    "delete": "Delete",
+    "status": "Status",
+    "confirmDeleteTitle": "Are you sure to delete?",
```
Jdallos, line 261:
```suggestion
    "confirmDeleteTitle": "Are you sure you want to delete?",
```

Jdallos: Bit wordy, probably confirm questionnaire delete is good for the title, and the button text can be delete

Jdallos (changes requested):

src/components/questionnaire/questionnaire-table.tsx
dalju005-xamk: Yep, this was left behind from when mockdata was rendered. I think this should be its own PR in the future. Where this status checking is added to this.

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+      } finally {
+        setLoading(false);
+      }
+    };
+    fetchQuestionnaires();
+  }, []);
+
+  const handleOpenDialog = (id: string, title: string) => {
+    setDeleteId(id);
+    setDeleteTitle(title);
```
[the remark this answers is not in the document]
dalju005-xamk: Hmm. This needs your help. Now when delete button is pressed (handleOpenDialog) it saves that questionnaires ID and Title to its own "temperary" state. Those states are then used in dialog where deleteQuestionnaire is called inside handleConfirmDelete(). So if approved the questionnaire that has saved in temporary state will be deleted.

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -18,59 +18,97 @@ import strings from "src/localization/strings";
...
+      } catch (error) {
+        setError(`${strings.error.questionnaireLoadFailed}, ${error}`);
       }
+      setLoading(false);
     };
     fetchQuestionnaires();
   }, []);
 
+  /**
+   * Function to open the dialog for deleting the questionnaire
```
Jdallos, line 53:
```suggestion
   * Function to open the dialog for deleting the questionnaire
   *
```

Jdallos: There should be a space between TSDoc description and the @param. Update when you notice it please- it's wrong in loads of places

src/components/questionnaire/questionnaire-table.tsx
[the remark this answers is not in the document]
Jdallos: But you also won't need to display the title of the questionnaire once it has been deleted, so it should still be able to be found in the questionnaires state until it is deleted and filtered out. Either way not a big deal and can be left as is.
Remember if you want me to see a comment you leave, don't resolve the conversation. I generally won't see a comment if you resolve the conversation

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -18,59 +18,97 @@ import strings from "src/localization/strings";
...
       );
       handleCloseDialog();
     } catch (error) {
-      console.error("Delete Questionnaire failed:", error);
-    } finally {
-      setLoading(false);
+      setError(`${strings.error.questionnaireDeleteFailed}, ${error}`);
     }
+      setLoading(false);
+
```
Jdallos, line 89:
```suggestion
```

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -18,59 +18,97 @@ import strings from "src/localization/strings";
...
     }
+      setLoading(false);
+    
+  };
+
+  /**
+   * Function to render the confirm delete dialog
+   */
+  const renderConfirmDeleteDialog = () => {
+    return (
```
Jdallos, line 96: This could be converted to an implicit return. e.g. remove the return () and replace the {} with () (or chat GPT if unsure).

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -102,12 +140,13 @@ const QuestionnaireTable = () => {
...
-            ) : (
-              <CloseIcon sx={{ color: "red" }} />
-            )
+          FIXME: "This should be checking if loggedInUser.id is in passedUsers array",
+          // renderCell: (params: GridRenderCellParams) =>
+          //   params.row.passedUsers && params.row.passedUsers.lenght > 0 ? (
+          //     <CheckCircleIcon sx={{ color: "green" }} />
+          //   ) : (
+          //     <CloseIcon sx={{ color: "red" }} />
+          //   )
```
Jdallos, line 149: What's going on here?

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -129,25 +168,11 @@ const QuestionnaireTable = () => {
           rows={questionnaires}
           columns={columns}
           pagination
-          pageSizeOptions={[5]}
-          getRowId={(row) => row.id}
+          getRowId={(row) => row.id || ""}
```
Jdallos, line 171: I could be wrong, but isn't this line unnecessary?

src/localization/en.json
```diff
@@ -29,7 +29,9 @@
     "dailyEntriesNotFound": "Could not find daily entries",
     "fetchFailedGeneral": "There was an error fetching data",
     "fetchFailedNoEntriesGeneral": "No time entries found",
-    "fetchSlackAvatarsFailed": "Error while fetching Slack avatars"
+    "fetchSlackAvatarsFailed": "Error while fetching Slack avatars",
+    "questionnaireLoadFailed": "Questionnaire loading failed",
+    "questionnaireDeleteFailed": "Questionnaire deletion failed"
```
Jdallos, line 34: Lets keep the error messages as consistant as possible, "Error while loading questionnaire", "Error while deleting questionnaire"

src/localization/en.json
```diff
@@ -258,7 +260,7 @@
     "edit": "Edit",
     "delete": "Delete",
     "status": "Status",
-    "confirmDeleteTitle": "Are you sure to delete?",
+    "confirmDeleteTitle": "Are you sure you want to delete?",
```
Jdallos, line 263: How about using the format string option to insert the name of the questionnaire to be deleted into this string, rather than having it separate in the Dialog content.
Message could be "Are you sure you want to delete questionnaire: [questionnarieTitle]?"

Or even shorter:
"Delete [questionnaireTitle]?"

You can see an example of how this is done in the timebank.atTheEndOf string. You will also need to look where in the code this is used to see the syntax for passing the title.

Jdallos: Would also need to be updated in the fi.json

Jdallos (approved):

PR #185, opened 2024-11-15
Some fixes to questionnaire UI

dalju005-xamk:


home-lambdas-API-spec
naStipe, line 1: Is this changed by pulling from GitHub?

naStipe (approved):
Seems fine to me

rimz178 (approved):
Looks good

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

rimz178 (approved):

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -171,10 +176,21 @@ const NewQuestionnaireScreen = () => {
...
           <NewQuestionnaireCard handleAddQuestion={handleAddQuestion} />
-          <CardActions
+          <Card
+          sx={{
+            p: 2,
+            mt: 4,
+            width: "100%",
+            display: "flex",
+            flexDirection: "column",
+            height: "100"
```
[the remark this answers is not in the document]
Jdallos: or just 100 without the quotes.

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -34,30 +35,35 @@ const NewQuestionnaireScreen = () => {
   const { questionnairesApi } = useLambdasApi();
   const [loading, setLoading] = useState(false);
   const setError = useSetAtom(errorAtom);
+  const [tooltipMessage, setTooltipMessage] = useState<string>("");
+  const [isDisabled, setIsDisabled] = useState<boolean>(true);
```
Jdallos, line 39: Rather than setting this to be true to begin with, why not set it based on the actual values we want to track for disabling e.g. where there is title and description?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -34,30 +35,35 @@ const NewQuestionnaireScreen = () => {
...
-    setQuestionnaire((prevQuestionnaire) => ({
-      ...prevQuestionnaire,
-      [name]: value
-    }));
+    setQuestionnaire((prevQuestionnaire) => {
+      const updatedQuestionnaire = {
+        ...prevQuestionnaire,
+        [name]: value,
+      } as Questionnaire;
+      updateTooltipAndIsDisabled();
```
Jdallos, line 59: Does this need to be called in the body of the set state? Can't it be called after the setQuestionnaire but still in the handleQuestionnaireInputChange fucntion?

src/components/screens/new-questionnaire-screen.tsx
Jdallos: In fact I think rather than having this isDisabled state you could reference the questionnaire.title || questionnaire.description in the disabled attribute of the button component where it is used.

I think this will save a lot of updating and tracking of the disabled state which maybe isn't necessary/ might be a little overcomplicated. What do you think?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -115,14 +121,41 @@ const NewQuestionnaireScreen = () => {
       questions: [],
       passScore: 0
     });
-    setLoading(false);
+    setTooltipMessage("");
+    setIsDisabled(true);
```
Jdallos, line 125: If you do remove the disabled state you could remove this

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -115,14 +121,41 @@ const NewQuestionnaireScreen = () => {
...
+  };
+
+  /**
+   * Function to check and set the tooltip message and isDisabled state
+   */
+  const updateTooltipAndIsDisabled = () => {
+    const isTitleEmpty = !questionnaire.title.trim();
+    const isDescriptionEmpty = !questionnaire.description.trim();
+    const isDisabled = isTitleEmpty || isDescriptionEmpty;
```
Jdallos, line 134: You wouldn't need this isDisabled state here if you remove as suggested

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -115,14 +121,41 @@ const NewQuestionnaireScreen = () => {
...
+    const tooltipMessage = isDisabled
+      ? isTitleEmpty && isDescriptionEmpty
+        ? strings.newQuestionnaireScreen.tooltipBothEmpty
+        : isTitleEmpty
+          ? strings.newQuestionnaireScreen.tooltipEmptyTitle
+          : strings.newQuestionnaireScreen.tooltipEmptyDescription
+      : "";
+
+    setTooltipMessage(tooltipMessage);
+    setIsDisabled(isDisabled);
```
Jdallos, line 145: No need here either

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -115,14 +121,41 @@ const NewQuestionnaireScreen = () => {
...
+    const isDescriptionEmpty = !questionnaire.description.trim();
+    const isDisabled = isTitleEmpty || isDescriptionEmpty;
+
+    const tooltipMessage = isDisabled
+      ? isTitleEmpty && isDescriptionEmpty
+        ? strings.newQuestionnaireScreen.tooltipBothEmpty
+        : isTitleEmpty
+          ? strings.newQuestionnaireScreen.tooltipEmptyTitle
+          : strings.newQuestionnaireScreen.tooltipEmptyDescription
+      : "";
```
Jdallos, line 142: Avoid nested ternaries. Maybe a switch or some if/if else conditions will be suitable here

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -115,14 +121,41 @@ const NewQuestionnaireScreen = () => {
...
   /**
    * Function to save the new questionnaire
    */
   const saveQuestionnaire = async () => {
     setLoading(true);
+
+    if (isDisabled) {
+      setLoading(false);
+      return;
+    }
```
Jdallos, line 157: This can be removed, the function can't be run unless the button isn't disabled anyway.

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -115,14 +121,41 @@ const NewQuestionnaireScreen = () => {
...
     });
-    setLoading(false);
+    setTooltipMessage("");
+    setIsDisabled(true);
+  };
+
+  /**
+   * Function to check and set the tooltip message and isDisabled state
+   */
+  const updateTooltipAndIsDisabled = () => {
```
Jdallos, line 131: Removing the disabled state you could rename this.

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -133,14 +166,18 @@ const NewQuestionnaireScreen = () => {
...
     } catch (error) {
       setError(`${strings.error.questionnaireSaveFailed}, ${error}`);
-      setLoading(false);
     }
+    setLoading(false);
   };
 
+  useEffect(() => {
+    updateTooltipAndIsDisabled();
+  }, [questionnaire.title, questionnaire.description]);
```
Jdallos, line 179: Rather than calling this in this UE and in the handleQuestionnaireInputChange, could it not be called directly from (and only from) the tooltip e.g. 
<Tooltip title={renderToolTip()}

This should work fine, you can see a very similar implementation here if it helps:
https://github.com/Metatavu/lipsanen-project-management-ui/pull/82/files

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -181,62 +218,65 @@ const NewQuestionnaireScreen = () => {
...
+              <Box sx={{ display: "flex", flexDirection: "column", width: "70%", mr: 4 }}>
+                <Typography
+                  variant="h6"
+                  gutterBottom
+                  sx={{ display: "flex", alignItems: "center", mb: 1, mt: 1 }}
+                >
+                  {strings.newQuestionnaireScreen.countedAnswers} {countCorrectAnswers()}
+                </Typography>
+                <Typography variant="h6" gutterBottom sx={{ mb: 1, mt: 1 }}>
+                  {strings.newQuestionnaireScreen.requiredAnswers} {questionnaire.passScore}
```
Jdallos, line 248: This could be done using the formatting strings option which allows us to pass the variable to the string. It's been done in other places in this project e.g. for the questionnaireTable.confirmDeleteTitle string.
Not essential in this case however

Jdallos (changes requested):

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -30,32 +31,37 @@ import { errorAtom } from "src/atoms/error";
...
    * @param event
    */
   const handleQuestionnaireInputChange = (event: ChangeEvent<HTMLInputElement>) => {
     const { name, value } = event.target;
-    setQuestionnaire((prevQuestionnaire) => ({
-      ...prevQuestionnaire,
-      [name]: value
-    }));
+
+    if (name === "title" || name === "description") {
```
Jdallos, line 54: Doesn't seem to me like it was necessary to add this condition in?

src/components/screens/new-questionnaire-screen.tsx
```diff
@@ -171,71 +199,75 @@ const NewQuestionnaireScreen = () => {
...
+              <Tooltip title={renderUpdatedTooltips()} placement="bottom">
+                <span>
+                  <Button
+                    sx={{ display: "flex", alignItems: "center", mt: 6, mr: 4 }}
+                    id="save-submit"
+                    size="large"
+                    variant="contained"
+                    color="success"
+                    onClick={saveQuestionnaire}
+                    disabled={loading || !questionnaire.title || !questionnaire.description}
```
Jdallos, line 256: Just pointing out these conditions could be placed into a named variable e.g. isDisabled just before the main component render. Not strictly necessary in this case but would be a good idea if the condition list grew any larger.

Jdallos (approved):

PR #191, opened 2024-12-02
Balance UI restoration

dalju005-xamk:
Restoration started, Users flextimes balances rendered in UI

### DIFF
--- a/src/components/home/balance-card.tsx
+++ b/src/components/home/balance-card.tsx
@@ -1,4 +1,4 @@
-import { Grid, Typography, Card, CardContent, Skeleton } from "@mui/material";
+import { Grid, Typography, Card, CardContent, Skeleton, Box } from "@mui/material";
 import strings from "src/localization/strings";
 import ScheduleIcon from "@mui/icons-material/Schedule";
 import { errorAtom } from "src/atoms/error";
@@ -32,24 +32,15 @@ const BalanceCard = () => {
    */
   const getUsersFlextimes = async () => {
     setLoading(true);
-    const loggedInUser = users.find(
-      (users: User) =>
-        users.id === userProfile?.id
-    );
-    console.log("loggedInUser", loggedInUser);
+    const loggedInUser = users.find((users: User) => users.id === userProfile?.id);
     if (loggedInUser) {
       const severaUserId = getSeveraUserId(loggedInUser);
-      console.log("severaUserId", severaUserId);
       try {
-        console.log(flexTimeApi);
         const fetchedUsersFlextime = await flexTimeApi.getFlextimeBySeveraUserId({
           severaUserId
         });
-        console.log("fetchedUsersFlextime:", fetchedUsersFlextime);
         setUsersFlextimes([fetchedUsersFlextime]);
-        console.log("usersFlextimes:", usersFlextimes);
       } catch (error) {
-        console.log(error)
         setError(`${strings.error.fetchFailedGeneral}, ${error}`);
       }
     }
@@ -57,7 +48,6 @@ const BalanceCard = () => {
   };
 
   useEffect(() => {
-    console.log("users", users);
     if (users.length > 0 && userProfile) {
       getUsersFlextimes();
     }
@@ -65,10 +55,18 @@ const BalanceCard = () => {
 
   const renderUserFlextime = () => {
     return (
-      <Typography variant="body1">
-        " No data"
-        FIXME: Localization
-      </Typography>
+      <Box>
+        {usersFlextimes?.map((flextime) => (
+          <Box key={flextime.totalFlextimeBalance} sx={{ mb: 2 }}>
+            <Typography variant="body1">
+              {strings.balanceCard.totalFlextimeBalance} {flextime.totalFlextimeBalance} hours
+            </Typography>
+            <Typography variant="body1">
+              {strings.balanceCard.monthFlextimeBalance} {flextime.monthFlextimeBalance} hours
+            </Typography>
+          </Box>
+        ))}
+      </Box>
     );
   };
 
--- a/src/components/home/balance-card.tsx
+++ b/src/components/home/balance-card.tsx
@@ -18,7 +18,7 @@ import { getSeveraUserId } from "src/utils/user-utils";
  * Component for displaying user's balance
  */
 const BalanceCard = () => {
-  const user = useAtomValue(usersAtom);
+  const users = useAtomValue(usersAtom);
   const userProfile = useAtomValue(userProfileAtom);
   const setError = useSetAtom(errorAtom);
   const [loading, setLoading] = useState(false);
@@ -32,7 +32,7 @@ const BalanceCard = () => {
    */
   const getUsersFlextimes = async () => {
     setLoading(true);
-    const loggedInUser = user.find((user: User) => user.id === userProfile?.id);
+    const loggedInUser = users.find((user: User) => user.id === userProfile?.id);
     if (loggedInUser) {
       const severaUserId = getSeveraUserId(loggedInUser);
       try {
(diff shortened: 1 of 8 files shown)


src/components/home/balance-card.tsx
```diff
@@ -1,91 +1,73 @@
...
-   * Renders person's total time
-   *
-   * @param personTotalTime PersonTotalTime
-   */
-  const renderPersonTotalTime = (personTotalTime: PersonTotalTime | undefined) => {
-    const balanceColor =
-      personTotalTime && personTotalTime.balance > 0
-        ? theme.palette.success.main
-        : theme.palette.error.main;
+  }, [users, userProfile]);
```
naStipe, line 55: Do we need TSDOCS here?

src/localization/strings.ts
```diff
@@ -360,6 +360,13 @@ export interface Localized extends LocalizedStringsMethods {
     cancel: string;
     confirm: string;
   };
```
naStipe, line 362: Maybe add empty line here

src/utils/user-utils.ts
```diff
@@ -0,0 +1,16 @@
+import type { User } from "src/generated/homeLambdasClient/models/User";
+import config from "src/app/config";
+
+/**
+ * To get the severa user id when logged in, if severaUserId is not found, it will return the test user severa id
+ * 
+ * @param user
```
naStipe, line 7: missing description (if needed)

src/components/home/balance-card.tsx
```diff
@@ -1,91 +1,73 @@
...
       } catch (error) {
         setError(`${strings.error.fetchFailedGeneral}, ${error}`);
       }
     }
     setLoading(false);
   };
 
-  /**
-   * Get person total time if it is undefined or set to "all time"
-   */
```
McDuuck, line 57: TSDocs for this maybe?

src/utils/user-utils.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @param user 
+ * @returns 
+ */
+export const getSeveraUserId = (user: User | undefined): string => {
+	if (user?.attributes?.severaUserId) {
+    return user.attributes.severaUserId;
+  }
+  const testUserSeveraId = config.user.testUserSeveraId || "";
+  return testUserSeveraId;
+};
```
naStipe, line 16: Is the formatting alright in here? That "if" seems to be in the wrong place

src/components/home/balance-card.tsx
```diff
@@ -1,91 +1,73 @@
...
-        <Typography color={balanceColor}>{strings.error.fetchFailedNoEntriesGeneral}</Typography>
-      );
-    }
-    if (personTotalTime) {
-      return (
-        <Typography color={balanceColor}>{getHoursAndMinutes(personTotalTime.balance)}</Typography>
-      );
-    }
-    return <Skeleton />;
+  const renderUserFlextime = () => {
```
McDuuck, line 56: And for this also?

src/utils/user-utils.ts
```diff
@@ -0,0 +1,16 @@
...
+import config from "src/app/config";
+
+/**
+ * To get the severa user id when logged in, if severaUserId is not found, it will return the test user severa id
+ * 
+ * @param user 
+ * @returns 
+ */
+export const getSeveraUserId = (user: User | undefined): string => {
+	if (user?.attributes?.severaUserId) {
```
McDuuck, line 11: intentation seems kinda off?

rimz178 (approved):

src/utils/user-utils.ts
dalju005-xamk: Simplified this a bit

naStipe (approved):

src/components/home/balance-card.tsx
```diff
@@ -1,50 +1,45 @@
...
-          personId: loggedInPerson?.id,
-          timespan: Timespan.ALL_TIME,
-          before: yesterday.toJSDate()
+        const fetchedUsersFlextime = await flexTimeApi.getFlextimeBySeveraUserId({
+          severaUserId
         });
-        setPersonTotalTime(fetchedPerson[0]);
+        setUsersFlextimes([fetchedUsersFlextime]);
       } catch (error) {
         setError(`${strings.error.fetchFailedGeneral}, ${error}`);
```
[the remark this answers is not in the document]
dalju005-xamk: Yes, This is left there from original BalanceCard.

src/utils/user-utils.ts
```diff
@@ -0,0 +1,12 @@
...
+import config from "src/app/config";
+
+/**
+ * To get the severa user id when logged in, if severaUserId is not found, it will return the test user severa id
+ *
+ * @param user
+ * @returns user severaUserId or testUserSeveraId
+ */
+export const getSeveraUserId = (user: User | undefined): string => {
+  return user?.attributes?.severaUserId ?? config.user.testUserSeveraId ?? "";
```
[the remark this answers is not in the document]
dalju005-xamk: This might need some thinking. We should not need this in production. As end users should have severaUserId. 
Do you mean that UI should give some message that user is using "test" data.

src/components/home/balance-card.tsx
```diff
@@ -1,91 +1,73 @@
...
+import type { User } from "src/generated/homeLambdasClient";
+import { useLambdasApi } from "src/hooks/use-api";
+import { getSeveraUserId } from "src/utils/user-utils";
 
 /**
  * Component for displaying user's balance
  */
 const BalanceCard = () => {
-  const persons = useAtomValue(personsAtom);
+  const user = useAtomValue(usersAtom);
```
Jdallos, line 21: this should be named users as it contains many users

src/components/home/balance-card.tsx
```diff
@@ -1,91 +1,73 @@
...
-        setError(`${strings.error.fetchFailedGeneral}, ${error}`);
+        setError(`${strings.error.fetchFailedFlextime}, ${error}`);
       }
     }
     setLoading(false);
   };
 
   /**
-   * Get person total time if it is undefined or set to "all time"
+   * Fetch user's flextime data when users or userProfile changes.
```
Jdallos, line 51: Is this accurate? Is userProfile being used in any way in this UE?

src/components/home/balance-card.tsx
```diff
@@ -1,91 +1,73 @@
...
-        <Typography color={balanceColor}>{getHoursAndMinutes(personTotalTime.balance)}</Typography>
-      );
-    }
-    return <Skeleton />;
+  const renderUserFlextime = () => {
+    return (
+      <Typography variant="body1">
+        {strings.formatString(
+          strings.balanceCard.totalFlextimeBalance,
+          usersFlextime?.totalFlextimeBalance ?? ""
```
Jdallos, line 67: If there is an error with fetching flex time an appropriate string should be renedered here rather than the current "Total flextime balance: hours"

src/utils/user-utils.ts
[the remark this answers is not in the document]
Jdallos: I mean currently if there is no user.attributes.severaUserId then it falls back to the testUserId, but it should not do this in production, rather it should render some error message informing user they do not have a serveraId linked to their account.

This will likely need some changes once we have the servera optIn keyword resolved.

At minumum I'd do a check for the environement, this is fine in develop condition but in production condition there should be an error string rendered if user does not have a serveraUserId. Probably will need to add some condition that the balance card does not render at all unless the user has severa optIn keyword (but this can be done in a future feature).

Jdallos (changes requested):

src/components/home/balance-card.tsx
```diff
@@ -47,29 +56,30 @@ const BalanceCard = () => {
...
-        )}
+        {strings.balanceCard.totalFlextimeBalance}{" "}
+        <span style={{ color: textColor }}>{totalFlextimeBalance}</span> {hourLabel}
       </Typography>
     );
   };
 
+  if (!severaUserId) {
+    return null;
+  }
```
Jdallos, line 81: I was hoping this could be controlled from the screens/home-screen like for the isDeveloper condition. 

You might need to check with Stephan to set up the severaID for your own user, for example granting yourself a severaId (probably the test one) and then seeing if it can be accessed through the access token in a similar way to the UserRoleUtils.isDeveloper

src/utils/user-utils.ts
```diff
@@ -1,12 +1,21 @@
...
  */
 export const getSeveraUserId = (user: User | undefined): string => {
-  return user?.attributes?.severaUserId ?? config.user.testUserSeveraId ?? "";
+  const severaUserId = user?.attributes?.severaUserId;
+
+  if (!severaUserId) {
+    if (import.meta.env.MODE === "development") {
+      return config.user.testUserSeveraId ?? "";
+    }
+    throw new Error(strings.error.noSeveraUserId);
```
Jdallos, line 18: Throwing the error like this seems to crash the whole app. I would suggest using the ErrorHandler like we do in other places, also consider in this case what should be rendered IN the balance card when there is no data and once the error handler has been closed. Perhaps a message "Something went wrong getting your balance"

Jdallos (changes requested):

Jdallos (approved):

PR #205, opened 2025-01-21
Fixing workHours fetching, type for phases, and mapping method for ro…

Hanhz2707:
…wsPhases

### DIFF
--- a/src/components/screens/sprint-view-screen.tsx
+++ b/src/components/screens/sprint-view-screen.tsx
@@ -156,7 +156,7 @@ const SprintViewScreen = () => {
           </Card>
           {resourceAllocations?.map((resourceAllocations) => (
             <TaskTable
-              key={resourceAllocations.phase?.severaPhaseId}
+              key={resourceAllocations.project?.severaProjectId}
               project={resourceAllocations.project ?? {} as ResourceAllocationsProject}
             />
           ))}
--- a/src/components/screens/sprint-view-screen.tsx
+++ b/src/components/screens/sprint-view-screen.tsx
@@ -1,3 +1,4 @@
+import { KeyboardReturn } from "@mui/icons-material";
 import {
   Box,
   Button,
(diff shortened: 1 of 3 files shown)


naStipe (approved):

McDuuck (approved):

src/components/screens/sprint-view-screen.tsx
```diff
@@ -156,8 +157,7 @@ const SprintViewScreen = () => {
           {resourceAllocations?.map((resourceAllocations) => (
             <TaskTable
               key={resourceAllocations.phase?.severaPhaseId}
```
Jdallos, line 159: Is there any reason not to pass the resourceAllocation.project.severaProjectId here? Not a big deal but as we don't seem to be passing the phase here but the project it maybe is a bit clearer?

src/components/screens/sprint-view-screen.tsx
```diff
@@ -1,3 +1,4 @@
+import { KeyboardReturn } from "@mui/icons-material";
```
Jdallos, line 1: Any idea why some of the imports are getting reordered in these reviews? I suspect it is related to the linter (biome) and someone not having it set up correctly. Perhaps something worth mentioning in the trainees daily to check everyones is working correctly.

src/components/sprint-view-table/tasks-table.tsx
```diff
@@ -8,35 +8,42 @@ import {
...
+  project: ResourceAllocationsProject
 }
 
 /**
  * Task table component
  *
  * @param props component properties
  */
-const TaskTable = ({ phases, filter, project }: Props) => {
+const TaskTable = ({filter, project }: Props) => {
```
Jdallos, line 41:
```suggestion
const TaskTable = ({ filter, project }: Props) => {
```

Jdallos (approved):

PR #206, opened 2025-01-21
Adding assignee for phase, redoing totalWorkHours logic

Hanhz2707:


src/utils/sprint-utils.ts
```diff
@@ -109,6 +109,36 @@ export const getSeveraUserId = (user: User): string => {
...
+/**
+ * Get phases's assignee
+ * 
+ * @param workHours with type WorkHours[]
+ * @param userId with type string
+ * 
+ * @returns assignee's name excludes given userId
+ */
+const getAssigneeWorkHours = (workHours: WorkHours[]) => {
+  const assignee = new Set(
```
McDuuck, line 136: Isnt "Set" used for only unique items, what if theres 2 person with the same name does this still work?

McDuuck: I might be wrong also, but it was first thing that came to mind with this one

Hanhz2707: Hmm actually ,that's good question. Maybe i will adding some filter for the ID as well. Good point !!!

rimz178 (approved):

McDuuck (approved):

src/utils/sprint-utils.ts
```diff
@@ -109,6 +109,40 @@ export const getSeveraUserId = (user: User): string => {
...
+ * 
+ * @param workHours with type WorkHours[]
+ * @param phase with type Phase
+ * @param userId with type string
+ * 
+ * @returns total work hours for each phase according to the userId
+ */
+const totalWorkHours = (workHours: WorkHours[], phase: Phase, userId: string) => {
+  return workHours
+    .filter(workHour => workHour.phase?.severaPhaseId === phase.severaPhaseId && workHour.user?.severaUserId === userId)
```
Jdallos, line 123: I think this could be simplified, or at least made clearer by having the phaseId comparison and userId comparison as named variables?

src/utils/sprint-utils.ts
```diff
@@ -117,16 +151,14 @@ export const getSeveraUserId = (user: User): string => {
  *
```
Jdallos, line 151: Missing userId @param

Jdallos (approved):

PR #207, opened 2025-01-21
feature-192-update-status-functionality

naStipe:


AfoninOleksandr (approved):
Looks good

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -200,6 +199,70 @@ const VacationRequestsScreen = () => {
...
+      );
+
+      setVacationRequests((prevRequests) =>
+        prevRequests.map((vacationRequest) => {
+          const updatedRequest = updatedVacationRequests.find(
+            (req) => req?.id === vacationRequest.id
+          );
+          return updatedRequest || vacationRequest;
+        })
+      );
```
Hanhz2707, line 259:
```suggestion
      setVacationRequests((prevRequests) =>
        prevRequests.map((vacationRequest) => {
          const updatedRequest = updatedVacationRequests.find(
            (req) => req?.id === vacationRequest.id
          );
          return updatedRequest || vacationRequest;
        })
      );
```setVacationRequests((prevRequests) =>
  prevRequests.map((vacationRequest) =>
    updatedVacationRequests.find((req) => req?.id === vacationRequest.id) || vacationRequest
  )
);

Comment (Hanhz2707) [inline src/components/screens/vacation-requests-screen.tsx:259]:
```suggestion
  setVacationRequests((prevRequests) =>
    prevRequests.map((vacationRequest) =>
      updatedVacationRequests.find((req) => req?.id === vacationRequest.id) || vacationRequest
    )
  );
```

Hanhz2707: This can be return directly instead of creating a new var then assign to it

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -200,6 +199,70 @@ const VacationRequestsScreen = () => {
...
+    if (!loggedInUser) return;
+
+    try {
+      setLoading(true);
+      const updatedVacationRequests = await Promise.all(
+        selectedRowIds.map(async (vacationRequestId) => {
+          const vacationRequest = vacationRequests.find(
+            (vacationRequest) => vacationRequest.id === vacationRequestId
+          );
+          if (!vacationRequest) return null;
```
Hanhz2707, line 221:
```suggestion
          if (!vacationRequest) return;

```

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -200,6 +199,70 @@ const VacationRequestsScreen = () => {
...
+            message: `This vacation request is ${status}`
+          };
+
+          const updatedStatus = statusExists
+            ? vacationRequest.status?.map((existingStatus) =>
+                existingStatus.createdBy === loggedInUser.id
+                  ? { ...existingStatus, ...newOrUpdatedStatus }
+                  : existingStatus
+              )
+            : [...(vacationRequest.status || []), newOrUpdatedStatus];
```
Hanhz2707, line 240: Are there any reasons why we need this brackets
```suggestion
            : [...vacationRequest.status || [], newOrUpdatedStatus];
```

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -200,6 +199,70 @@ const VacationRequestsScreen = () => {
...
+
+          return await vacationRequestsApi.updateVacationRequest({
```
Hanhz2707, line 242: If we already have Promise.all above, maybe we dont need await here as Promise.all should wait for all promises to resolve it
```suggestion
          return vacationRequestsApi.updateVacationRequest({
```

src/components/vacation-requests-table/vacation-requests-table.tsx
```diff
@@ -109,6 +112,7 @@ const VacationRequestsTable = ({
 
         if (vacationRequest.message.length) {
           row.message = vacationRequest.message;
+          row.message = vacationRequest.message;
         }
```
Hanhz2707, line 116: Is this duplication on purposes ?

Hanhz2707 (approved):
Approved with minor suggestions

.gitignore
```diff
@@ -34,7 +34,6 @@ lerna-debug.log*
 .vscode/*
 !.vscode/extensions.json
 .idea
-.DS_Store
```
Jdallos, line 37: I can't think of why this would have been added as I don't think we have had any mac users  in the team. Still, are you sure it needs to be removed?

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -67,35 +67,35 @@ const VacationRequestsScreen = () => {
...
+    } catch (error) {
+      setError(`${strings.vacationRequestError.fetchRequestError}, ${error}`);
     }
     setLoading(false);
   };
 
-  useMemo(() => {
+  /**
+   * Re-fetch vacation requests whenever the component is mounted
+   */
```
Jdallos, line 90: Typically we don't add docs the Use Effects, and I guess while this description is correct, it's also incomplete. I'd suggest removing

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -200,6 +199,67 @@ const VacationRequestsScreen = () => {
...
+
+          const statusExists = vacationRequest.status?.some(
+            (existingStatus) => existingStatus.createdBy === loggedInUser.id
+          );
+
+          const newOrUpdatedStatus = {
+            status,
+            createdBy: loggedInUser.id,
+            updatedAt: new Date(),
+            message: `This vacation request is ${status}`
```
Jdallos, line 231: This seems a bit redundant, we we have some further development plans for it e.g. custom text input from the status setter? Otherwise is it needed?

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -200,6 +199,67 @@ const VacationRequestsScreen = () => {
...
+          };
+
+          const updatedStatus = statusExists
+            ? vacationRequest.status?.map((existingStatus) =>
+                existingStatus.createdBy === loggedInUser.id
+                  ? { ...existingStatus, ...newOrUpdatedStatus }
+                  : existingStatus
+              )
+            : [...vacationRequest.status || [], newOrUpdatedStatus];
```
Jdallos, line 240: General rule is to avoid nested ternary, which can sometimes be annoying i know!
Small refactor here to get rid of the nested ternary and improve clarity

Jdallos: The || [] seems a little odd, are you sure it's needed?

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -200,6 +199,67 @@ const VacationRequestsScreen = () => {
...
+            createdBy: loggedInUser.id,
+            updatedAt: new Date(),
+            message: `This vacation request is ${status}`
+          };
+
+          const updatedStatus = statusExists
+            ? vacationRequest.status?.map((existingStatus) =>
+                existingStatus.createdBy === loggedInUser.id
+                  ? { ...existingStatus, ...newOrUpdatedStatus }
+                  : existingStatus
```
Jdallos, line 238: If existingStatus is true don't we know that it was created by the logged in user? In which case this inner ternary becomes redundant?

Also If there is an existing status from the current user don't we want to update it/ replace it, rather than just add another status on top of it { ...existingStatus, ...newOrUpdatedStatus }? And in the Or case, if the existing status wasn't created by the currently updating user we are now replacing it : existingStatus, so replacing a status created by someone else. This also doesn't seem right. Perhaps these conditions are the wrong way round? Perhaps I'm mistaken, lets discuss further if needed

Jdallos (changes requested):

.gitignore
naStipe: I did not do it manually, so not sure why it is gone, but I will put it back just in case

src/components/screens/vacation-requests-screen.tsx
naStipe: Is the message needed? I think I added it just because we used to have it with timebank, if that`s what you`re asking

.gitignore
naStipe: It is defined twice in the document

src/components/screens/vacation-requests-screen.tsx
naStipe: || [] ensures that an array is returned, it gives out an error without it,

naStipe: I believe that the nested ternary is actually necessary in here. First one is needed, because we need to decide on action depending on whether the statusExists is true or not. The second one is needed for mapping method to work, as far as I understood. I did not see a way to map without it.

Jdallos (approved):

PR #224, opened 2025-06-04
Feature 222 questionnaire passed score card

AYZBTR:
Display user's questionnaire completion status (X out of Y passed) with clickable navigation to questionnaire section.

Bugfix: prevent state updates on unmounted SprintViewCardContent component, Add cleanup logic with isMountedRef to avoid errors when navigating back with browser button.
key changes made for bug fix:

1. Added isMountedRef to track if the component is still mounted
2. Added cleanup in useEffect to set isMountedRef.current = false when component unmounts
3. Check isMountedRef.current before updating state in the async function
4. Improved error handling to ignore aborted requests (which happen during navigation)
5. Added early return in getAllocationsAndProjects if data is already loaded

This should prevent the error when navigating back with the browser button. The component will now properly handle being unmounted during an active API request and won't try to update state after unmounting

### DIFF
--- a/src/components/home/questionnaire-card.tsx
+++ b/src/components/home/questionnaire-card.tsx
@@ -19,7 +19,7 @@ const QuestionnaireCard = () => {
       return (
         <CardContent>
           <Typography variant="h6" fontWeight={"bold"} style={{ marginTop: 6, marginBottom: 3 }}>
-            {strings.questionnaireCard?.questionnairesBuilder || "Questionnaires builder"}
+            {strings.questionnaireCard?.questionnairesBuilder}
           </Typography>
         </CardContent>
       );
@@ -28,7 +28,7 @@ const QuestionnaireCard = () => {
     return (
       <CardContent>
         <Typography variant="h6" fontWeight={"bold"} style={{ marginTop: 6, marginBottom: 3 }}>
-          {strings.questionnaireProgress?.title || "Questionnaires"}
+          {strings.questionnaireProgress?.title}
         </Typography>
         <Grid container>
           <Box sx={{ width: "100%", display: "flex", flexDirection: "column" }}>
--- a/src/components/home/questionnaire-card.tsx
+++ b/src/components/home/questionnaire-card.tsx
@@ -1,7 +1,8 @@
-import { Typography, Card, CardContent, Grid } from "@mui/material";
+import { Typography, Card, CardContent, Grid, Box } from "@mui/material";
 import { Link } from "react-router-dom";
 import strings from "src/localization/strings";
 import UserRoleUtils from "src/utils/user-role-utils";
+import QuestionnaireProgress from "./questionnaire-progress";
 
 /**
  * Component for displaying questionnaire card
@@ -18,7 +19,7 @@ const QuestionnaireCard = () => {
       return (
         <CardContent>
           <Typography variant="h6" fontWeight={"bold"} style={{ marginTop: 6, marginBottom: 3 }}>
-            {strings.questionnaireCard.questionnairesBuilder}
+            {strings.questionnaireCard?.questionnairesBuilder || "Questionnaires builder"}
           </Typography>
         </CardContent>
       );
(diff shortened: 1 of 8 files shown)


balashova12 (approved):
approved

src/components/home/questionnaire-progress.tsx
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Component to display user's questionnaire progress on home screen
+ * Located in home folder as it's specific to the dashboard
+ */
+const QuestionnaireProgress = () => {
+  const { questionnairesApi } = useLambdasApi();
+  const userProfile = useAtomValue(userProfileAtom);
+  const users = useAtomValue(usersAtom);
+  const [questionnaires, setQuestionnaires] = useState<Questionnaire[]>([]);
+  const [loading, setLoading] = useState(true);
```
Jdallos, line 19: We initialize loading as false and only set it to true when we need to  (when making  request).

Jdallos: So this should be set to false

src/components/screens/home-screen.tsx
```diff
@@ -54,11 +57,39 @@ const HomeScreen = () => {
...
+          </Typography>
+          <Grid container>
+            <Box sx={{ width: "100%", display: "flex", flexDirection: "column" }}>
+              <QuestionnaireProgress />
+            </Box>
+          </Grid>
+        </CardContent>
+      </Card>
+    </Link>
+  );
```
Jdallos, line 89: We already have the QuestionnaireCard, shouldn't this be done in there rather than here. All other cards render the card content in the file related to the card.

src/components/home/sprint-view-card-content/user-sprint-view-card.tsx
```diff
@@ -31,7 +31,16 @@ const SprintViewCardContent = () => {
   const [resourceAllocations, setResourceAllocations] = useState<ResourceAllocations[]>([]);
   const { resourceAllocationsApi } = useLambdasApi();
   const setError = useSetAtom(errorAtom);
-  const filteredAllocations = filterAllocations(resourceAllocations, adminMode)
+  const filteredAllocations = filterAllocations(resourceAllocations, adminMode);
+  const isMountedRef = useRef(true);
+
+  useEffect(() => {
```
Jdallos, line 37: All changes in this file seem unrelated to the questionnaires progress feature. Can you explain what these changes are for? Otherwise please undo these changes.

Jdallos (changes requested):

src/components/home/sprint-view-card-content/user-sprint-view-card.tsx
AYZBTR: The changes in this file are part of the browser navigation bug fix, not the questionnaire progress feature itself.

When users clicked the questionnaire card and navigated back using the browser's back button, the SprintViewCardContent component was throwing a 500 error because it was trying to update state after being unmounted. 

The fix adds:
- `isMountedRef` to track component lifecycle
- Cleanup logic in useEffect to set the ref to false on unmount
- Guards before all setState calls to check if component is still mounted

This prevents the "Failed to load resource: the server responded with a status of 500" error that was appearing when navigating back from the questionnaire page. While not directly part of the questionnaire feature, this fix was necessary to ensure smooth navigation between the home dashboard (where both questionnaire progress and sprint view are displayed) and the questionnaire section.

src/components/home/questionnaire-card.tsx
```diff
@@ -27,12 +28,12 @@ const QuestionnaireCard = () => {
     return (
       <CardContent>
         <Typography variant="h6" fontWeight={"bold"} style={{ marginTop: 6, marginBottom: 3 }}>
-          {strings.questionnaireCard.questionnaires}
+          {strings.questionnaireProgress?.title}
```
Jdallos, line 31: This is still using optional syntax- the questionnaireProgress?.title- the ?.

Is there a reason for this, otherwise it should be removed.

src/components/home/questionnaire-card.tsx
```diff
@@ -18,7 +19,7 @@ const QuestionnaireCard = () => {
       return (
         <CardContent>
           <Typography variant="h6" fontWeight={"bold"} style={{ marginTop: 6, marginBottom: 3 }}>
-            {strings.questionnaireCard.questionnairesBuilder}
+            {strings.questionnaireCard?.questionnairesBuilder}
```
Jdallos, line 22: This is still using optional syntax- the questionnaireCard?.questionnairesBuilder- the ?.

Is there a reason for this, otherwise it should be removed.

PR #255, opened 2025-09-02
feature-251-form-bugs

JuusoHeikkinen02:
Added Tooltip to new-question-card submit button and have it disabled if required fields are not filled. 

Added validation to adding new software in software registry and disabled the submit button if required fields are not filled. 

Added "add tag" button to the addSoftwareModal and limited shown tags to only show the first three and after that load chip that has the count of the remaining hidden tags. 

For the Tooltip had to create separate function to check ValidationConditions so a correct Tooltip message is shown.

### DIFF
--- a/src/components/questionnaire/new-question-card.tsx
+++ b/src/components/questionnaire/new-question-card.tsx
@@ -13,7 +13,10 @@ import type React from "react";
 import { useState } from "react";
 import type { AnswerOption } from "src/generated/homeLambdasClient";
 import strings from "src/localization/strings";
-import { isQuestionValid, getQuestionValidationTooltipMessage} from "src/utils/questionnaireBuilderUtils";
+import {
+  isQuestionValid,
+  getQuestionValidationTooltipMessage
+} from "src/utils/questionnaireBuilderUtils";
 
 /**
  * Interface for the NewQuestionCard component
@@ -163,22 +166,22 @@ const NewQuestionCard = ({ handleAddQuestion }: Props) => {
                 {strings.newQuestionnaireCard.addAnswer}
               </Typography>
             </Button>
-            <Tooltip 
-                title={getQuestionValidationTooltipMessage({questionText, answerOptions}, strings)} 
-                placement="bottom"
-                disableHoverListener={isQuestionValid({ questionText, answerOptions })}
-              >
-            <span>
-            <Button
-              sx={{ alignItems: "center" }}
-              size="large"
-              variant="contained"
-              onClick={handleAddNewQuestion}
-              disabled={isDisabled}
+            <Tooltip
+              title={getQuestionValidationTooltipMessage({ questionText, answerOptions }, strings)}
+              placement="bottom"
+              disableHoverListener={isQuestionValid({ questionText, answerOptions })}
             >
-              {strings.newQuestionnaireCard.saveAnswer}
-            </Button>
-            </span>
+              <Box>
+                <Button
+                  sx={{ alignItems: "center" }}
+                  size="large"
+                  variant="contained"
+                  onClick={handleAddNewQuestion}
+                  disabled={isDisabled}
+                >
+                  {strings.newQuestionnaireCard.saveAnswer}
+                </Button>
+              </Box>
             </Tooltip>
           </CardActions>
         </CardContent>
--- a/src/components/software-registry/SoftwareDetails.tsx
+++ b/src/components/software-registry/SoftwareDetails.tsx
@@ -32,7 +32,7 @@ const SoftwareDetails: FunctionComponent = () => {
   const [loading, setLoading] = useState(false);
   const [software, setSoftware] = useState<SoftwareRegistry | null>(null);
   const [softwareList, setSoftwareList] = useAtom(softwareAtom);
-  const [error, setError] = useState<string | null>(null);
+  const [error, setError] = useState<string |  { message: string; details?: any } | null>(null);
   const [createdByUserName, setCreatedByUserName] = useState<string>("");
   const [isEditModalOpen, setIsEditModalOpen] = useState(false);
   const navigate = useNavigate();
@@ -47,9 +47,6 @@ const SoftwareDetails: FunctionComponent = () => {
     fetchSoftwareDetails();
   }, [id]);
 
-  /**
-   * Fetches all software to list
-   */
   useEffect(() => {
     fetchSoftwaresToList();
   }, []);
@@ -75,7 +72,7 @@ const SoftwareDetails: FunctionComponent = () => {
   };
 
   /**
-   * This function retrieves the list of software applications from the API and updates the state.
+   * List software and update the state.
    */
   const fetchSoftwaresToList = async () => {
     setLoading(true);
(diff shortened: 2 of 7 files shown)


src/utils/questionnaireBuilderUtils.ts
```diff
@@ -280,6 +283,17 @@ export const isFormValid = (questionnaire: Questionnaire): boolean => {
...
 
+export const isQuestionValid = (question: Question): boolean => {
+  if (!question.questionText || 
+      question.questionText.trim() === "" || 
+      !question.answerOptions || question.answerOptions.length === 0 || 
+      !question.answerOptions.every(option=>option.label.trim() !=="")) {
+    return false;
+  }
+  return true;
+
```
StudVladimir, line 294: remove space

src/utils/questionnaireBuilderUtils.ts
```diff
@@ -325,3 +339,74 @@ export const getValidationTooltipMessage = (
...
+    case ValidationCondition.NO_QUESTIONS:
+      return strings.error.questionnaireSaveFailed;
+    case ValidationCondition.QUESTIONS_AND_ANSWERS_EMPTY:
+      return strings.newQuestionnaireCard.tooltipBothEmpty;
+    case ValidationCondition.QUESTIONS_EMPTY:
+    return strings.newQuestionnaireCard.tooltipEmptyQuestion;
+    case ValidationCondition.ANSWERS_EMPTY:
+    return strings.newQuestionnaireCard.tooltipEmptyAnswers;
+    default:
+      return strings.error.generic;
```
StudVladimir, line 408: remove space at the top line, and make tabulation in start of each line looks the same

src/components/questionnaire/new-question-card.tsx
```diff
@@ -159,14 +163,23 @@ const NewQuestionCard = ({ handleAddQuestion }: Props) => {
...
             <Button
               sx={{ alignItems: "center" }}
               size="large"
               variant="contained"
               onClick={handleAddNewQuestion}
+              disabled={isDisabled}
             >
               {strings.newQuestionnaireCard.saveAnswer}
             </Button>
+            </span>
```
Jdallos, line 181: Double check the indentation between span and button. And do we really need a span wrapping the button? I'd suggest removing

src/components/software-registry/SoftwareDetails.tsx
```diff
@@ -45,6 +47,13 @@ const SoftwareDetails: FunctionComponent = () => {
     fetchSoftwareDetails();
   }, [id]);
 
+  /**
+   * Fetches all software to list
+   */
```
Jdallos, line 52: Unnecessary comment

src/components/software-registry/SoftwareDetails.tsx
```diff
@@ -65,14 +74,29 @@ const SoftwareDetails: FunctionComponent = () => {
     }
   };
 
+  /**
+   * This function retrieves the list of software applications from the API and updates the state.
```
Jdallos, line 78:
```suggestion
   * List software and update the state.
```
It's good to keep the comments short where possible

src/components/software-registry/SoftwareDetails.tsx
```diff
@@ -65,14 +74,29 @@ const SoftwareDetails: FunctionComponent = () => {
...
+   */
+  const fetchSoftwaresToList = async () => {
+    setLoading(true);
+    try {
+      const fetchedSoftware = await softwareApi.listSoftware();
+      setSoftwareList(fetchedSoftware);
+    } catch (error) {
+      setError(`Error fetching software data: ${error}`);
```
Jdallos, line 86: Localize error

src/components/software-registry/SoftwareDetails.tsx
```diff
@@ -194,12 +218,12 @@ const SoftwareDetails: FunctionComponent = () => {
...
         {software.image && (
-          <img src={software.image}
+          <img
+            src={software.image}
             alt={software.name}
             style={{ width: "200px", height: "200px", objectFit: "contain" }} />
         )}
-        <Typography gutterBottom
-        variant="h3"
+        <Typography gutterBottom variant="h3"
```
Jdallos, line 226: Your biome is clearly working now, nice!

src/utils/questionnaireBuilderUtils.ts
```diff
@@ -204,74 +192,80 @@ export enum ValidationCondition {
   TITLE_EMPTY = "titleEmpty",
   DESCRIPTION_EMPTY = "descriptionEmpty",
   NO_QUESTIONS = "noQuestions",
-  NO_CORRECT_ANSWERS = "noCorrectAnswers", 
+  QUESTIONS_AND_ANSWERS_EMPTY = "questionsAndAnswersEmpty",
+  QUESTIONS_EMPTY = "emptyQuestions",
```
Jdallos, line 196: It seems to add an unnecessary layer to me that we have this enum to handle validation conditions and then later convert it to a string. Why not just set it as a string here straight away?
It might be that it causes issues to use the localization library hook in a .ts file, if so then this is fine as it is.

src/utils/questionnaireBuilderUtils.ts
```diff
@@ -280,16 +274,29 @@ export const isFormValid = (questionnaire: Questionnaire): boolean => {
   return isValid;
 };
 
+export const isQuestionValid = (question: Question): boolean => {
```
Jdallos, line 277: Missing TSDocs

src/utils/questionnaireBuilderUtils.ts
```diff
@@ -325,3 +332,76 @@ export const getValidationTooltipMessage = (
       return strings.error.generic;
   }
 };
+
+export const validateQuestion = (
```
Jdallos, line 336: Missing TSDocs

src/utils/questionnaireBuilderUtils.ts
```diff
@@ -325,3 +332,76 @@ export const getValidationTooltipMessage = (
...
+    return {
+      isValid: false,
+      condition: ValidationCondition.ANSWERS_EMPTY
+    };
+  }
+
+  return {
+    isValid: true,
+    condition: ValidationCondition.VALID
+  };
```
Jdallos, line 372: Could be a switch but not much difference in it

src/utils/questionnaireBuilderUtils.ts
```diff
@@ -325,3 +332,76 @@ export const getValidationTooltipMessage = (
...
+};
+
+export const getQuestionValidationTooltipMessage = (
```
Jdallos, line 375: MIssing TSDocs

src/utils/questionnaireBuilderUtils.ts
Jdallos: well we can access the strings in getQuestionValidationTooltipMessage in this file so don't see why we can't here. 
I could be mistaken but it seems to me setting the string rather than an enum here simplifies things

Jdallos (approved):

src/components/questionnaire/new-question-card.tsx
JuusoHeikkinen02: I used span because the tooltip has to have a single child that can hold ref. I just copied the method from new-question-builder.tsx where there is a span also used on the save button. Do I remove the span and make a custom button that can hold ref?

PR #273, opened 2025-10-01
Feature 270 change general global rules for design

JuusoHeikkinen02:
Global theme styling caused issues with the styling on the oncall screen, made a separate theme for software screens.

### DIFF
--- a/src/components/screens/all-software-screen.tsx
+++ b/src/components/screens/all-software-screen.tsx
@@ -21,7 +21,6 @@ import {
   DialogContent,
   DialogContentText,
   DialogTitle,
-  ThemeProvider
 } from "@mui/material";
 import Content from "../software-registry/allContent";
 import { useLambdasApi } from "src/hooks/use-api";
@@ -38,7 +37,6 @@ import SearchIcon from "@mui/icons-material/Search";
 import useCreateSoftware from "src/hooks/use-create-software";
 import AddSoftwareModal from "../software-registry/AddSoftwareModal";
 import BackButton from "../generics/back-button";
-import { SoftwareScreenTheme } from "src/theme";
 
 /**
  * All software screen component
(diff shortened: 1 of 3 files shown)


Alexander-K-byte (approved):

StudVladimir (approved):

src/components/screens/all-software-screen.tsx
```diff
@@ -289,6 +291,7 @@ const AllSoftwareScreen = () => {
   }
 
   return (
+    <ThemeProvider theme={SoftwareScreenTheme}>
```
Jdallos, line 294: Is this really the way to update the theme here?
Don't we already have the ThemeProvier in App.tsx and shouldn't we update the theme from there?

Jdallos: Although it looks like this is supported and so we can leave it like this if you think it's the best plan going forward.

src/components/screens/software-registry-screen.tsx
```diff
@@ -186,163 +193,148 @@ const SoftwareScreen = () => {
...
-              }}
-            >
-              <ArrowDownwardIcon />
-            </IconButton>
-          </Grid>
-        )}
-        <Grid item container justifyContent="space-between" alignItems="center">
-          <Typography variant="h3">
-            {strings.softwareRegistry.myApplications}
+    <ThemeProvider theme={SoftwareScreenTheme}>
```
Jdallos, line 196: But do we really need the same provider again here? Can't this inherit from the previous?

Jdallos (approved):

src/components/screens/all-software-screen.tsx
JuusoHeikkinen02: The problem was that we had a global select component style but it doesn't really work with oncall page, so i had to only wrap this theme to these two specific screens. The themeprovider inside the app.tsx provides the global theme for all components but I can't import these styles through it because then the oncall page has wrong style. I'm not sure if its the best way, but if its good enough I can apply these changes

PR #296, opened 2025-10-16
bugfix-290-fix-on-board-card-visible-without-permisson 

MarcelMeta:
also adjusted border size so all cards fit

### DIFF
--- a/src/components/onboarding/Onboarding.tsx
+++ b/src/components/onboarding/Onboarding.tsx
@@ -3,6 +3,7 @@ import { Box, Paper, Typography, Button, IconButton } from "@mui/material";
 import CloseIcon from "@mui/icons-material/Close";
 import { getOnboardingSteps } from "./onboardingSteps";
 import strings from "src/localization/strings";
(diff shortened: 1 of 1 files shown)


jessesillman (approved):

JuusoHeikkinen02 (approved):

src/components/onboarding/Onboarding.tsx
```diff
@@ -17,60 +18,101 @@ const Onboarding: React.FC = () => {
   const ONBOARDING_KEY = "onboardingComplete";
   const [targetRect, setTargetRect] = useState<DOMRect | null>(null);
   const rafRef = useRef<number | null>(null);
+  const { isAdmin, isDeveloper, isTester, isAccountant } = useUserRole();
 
-  // Always get the latest steps for the current language
+  /**
+   * Whether the current user has permission to view onboarding.
```
Jdallos: As a general rule we only use TSDOcs for function comments. But this is fine here, typically if we wanted to comment something other than a function we would use // syntax

PR #300, opened 2025-10-20
Feature 292 vacations consume user vacation days

JuusoHeikkinen02:
Added validation to create/draft/update functions of the vacationrequests so user gets error if trying to request a vacation that consumes more days than the user has unspentVacationdays. Moved the getDays function outside the useRow so it can be used elsewhere.

### DIFF
--- a/src/components/screens/vacation-requests-screen.tsx
+++ b/src/components/screens/vacation-requests-screen.tsx
@@ -21,7 +21,6 @@ import type { FilterType } from "src/utils/vacation-filter-type";
 import { validateUserVacationRequest } from "src/utils/vacations-utils";
 import BackButton from "../generics/back-button";
 import VacationRequestsTable from "../vacation-requests-table/vacation-requests-table";
-import { getDays } from "./admin-vacation-management/UserRow";
 
 /**
  * Vacation requests screen


src/components/screens/vacation-requests-screen.tsx
```diff
@@ -233,6 +247,16 @@ const VacationRequestsScreen = () => {
...
+      const unspentDays = Number(
+        getDays(loggedInUser.attributes?.unspentVacationDaysByYear, currentYear)
+      );
+      const requestedDays = Number(vacationRequestData.days ?? 0);
+      const { valid, errorMessage } = validateVacationRequestDays(requestedDays, unspentDays);
+      if (!valid) {
+        setError(errorMessage);
+        setLoading(false);
+        return;
+      }
```
Jdallos: Why not extract this into a named reusable function/ utility as it's used multiple times.

src/utils/vacations-utils.ts
```diff
@@ -57,17 +54,36 @@ export function parseVacationDays(user: User): VacationDays {
...
+export const validateVacationRequestDays = (
+  requestedDays: number,
+  unspentDays: number
+): { valid: boolean; errorMessage?: string } => {
+  if (requestedDays > unspentDays) {
+    return {
+      valid: false,
+      errorMessage: strings.vacationRequestError.tooManyDaysRequested
+        .replace("{requestedDays}", String(requestedDays))
+        .replace("{unspentDays}", String(unspentDays))
```
Jdallos: I think there is a cleaner way to pass variables into the strings without using the .replace. The curley brack syntax in the string is correct. Might be worth double checking

Jdallos (approved):

PR #324, opened 2025-11-19
Feature-320-Severa-Opt-Out

jessesillman:
## What has been done
- Implemented the toggle logic in setting-screen.tsx to revoke Severa opt-in.
- revokeSeveraOptIn calls the generated removeSeveraOptIn (**DELETE** _/users/{userId}/severa-opt-in_) and removes the the attributes from user atoms.

Note: regenerate open api spec for testing if needed.

### DIFF
--- a/src/components/screens/settings-screen.tsx
+++ b/src/components/screens/settings-screen.tsx
@@ -3,6 +3,7 @@ import { useAtom, useSetAtom } from "jotai";
 import { useEffect, useState } from "react";
 import { userProfileAtom } from "src/atoms/auth";
 import { errorAtom } from "src/atoms/error";
+import { usersAtom } from "src/atoms/user";
 import { useLambdasApi } from "src/hooks/use-api";
 import strings from "src/localization/strings";
 
@@ -10,8 +11,9 @@ import strings from "src/localization/strings";
  * Settings screen component
  */
 const SettingsScreen = () => {
+  const [userProfile, setUserProfile] = useAtom(userProfileAtom);
   const { usersApi } = useLambdasApi();
-  const [userProfile] = useAtom(userProfileAtom);
+  const setUsers = useSetAtom(usersAtom);
   const setError = useSetAtom(errorAtom);
 
   const [isConsentGiven, setIsConsentGiven] = useState<boolean>(
@@ -22,7 +24,7 @@ const SettingsScreen = () => {
   useEffect(() => {
     setIsConsentGiven(Boolean(userProfile?.attributes?.severaUserId));
   }, [userProfile?.attributes?.severaUserId]);
-  
+
   /**
    * Handles toggle change event
    */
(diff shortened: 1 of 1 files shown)


JuusoHeikkinen02 (approved):

src/components/screens/settings-screen.tsx
```diff
@@ -16,54 +16,95 @@ const SettingsScreen = () => {
...
     Boolean(userProfile?.attributes?.severaUserId)
   );
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  useEffect(() => {
+    setIsConsentGiven(Boolean(userProfile?.attributes?.severaUserId));
+  }, [userProfile?.attributes?.severaUserId]);
 
   const handleToggleChange = () => {
```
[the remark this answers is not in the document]
Jdallos: Even if the function names are quite self explanatary we should still add the TSDoc comments, a few workds/ single line is sufficient

src/components/screens/settings-screen.tsx
```diff
@@ -16,54 +16,95 @@ const SettingsScreen = () => {
...
+    }
   };
 
   const grantSeveraOptInConsent = async () => {
     setIsLoading(true);
     try {
+      if (!usersApi) {
+        setError(strings.error.fetchFailedFlextime);
+        return;
+      }
```
Jdallos: Do we really need to check if the usersApi exists? It should, so I'm not sure why this is necessary

src/components/screens/settings-screen.tsx
```diff
@@ -16,54 +16,95 @@ const SettingsScreen = () => {
...
-          prev.map((u) =>
-            u.id === userProfile.id ? { ...u, attributes: { ...u.attributes, severaUserId } } : u
-          )
+          prev.map((u) => (u.id === userProfile.id ? { ...u, attributes: updatedAttributes } : u))
         );
       }
     } catch (error) {
-      console.error("Error fetching consent:", error);
-      setError(`${strings.error.fetchFailedFlextime}, ${error}`);
+      setError(`${strings.error.fetchFailedFlextime}, ${String(error)}`);
```
Jdallos: Error string is incorrect, this is not related to flex time.

src/components/screens/settings-screen.tsx
```diff
@@ -16,54 +16,95 @@ const SettingsScreen = () => {
...
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const revokeSeveraOptIn = async () => {
```
Jdallos: TSDOcs

src/components/screens/settings-screen.tsx
```diff
@@ -16,54 +16,95 @@ const SettingsScreen = () => {
...
+        string[] | string | undefined
+      >;
+      delete updatedAttributes.severaUserId;
+      delete updatedAttributes.isSeveraOptIn;
+
+      const updatedProfile = { ...userProfile, attributes: updatedAttributes };
+      setUserProfile(updatedProfile);
+      setUsers((prev) =>
+        prev.map((u) => (u.id === userProfile.id ? { ...u, attributes: updatedAttributes } : u))
+      );
```
Jdallos: As said in a previous comment if this is a state patch to reflect the keycloak updates without the user having to logout then it's fine, but I wonder if there is a cleaner way to just refetch the keycloak data...

src/components/screens/settings-screen.tsx
```diff
@@ -16,54 +16,95 @@ const SettingsScreen = () => {
...
+
+      setIsConsentGiven(false);
+    } catch (error) {
+      setError(`${strings.error.fetchFailedFlextime}, ${String(error)}`);
```
Jdallos: This error does not seem to be the right string for this catch- this is not fetching flextime

src/components/screens/settings-screen.tsx
```diff
@@ -12,60 +11,72 @@ import strings from "src/localization/strings";
...
-            severaUserId
-          }
-        };
-
-        setUserProfile(updatedProfile);
-        setUsers((prev) =>
-          prev.map((u) =>
-            u.id === userProfile.id ? { ...u, attributes: { ...u.attributes, severaUserId } } : u
-          )
-        );
```
Jdallos: So now if you have removed the profile updating code will the keycloak atom/ keycloak profile actually be updated straight away for the user or will they need to logout and log back in (which would be very bad UI)

Jdallos (approved):

src/components/screens/settings-screen.tsx
jessesillman: Right, removing this part might've caused an issue with immediate update when toggled out. Before merging this I will try to find another solution for this update part while keeping the code clean / putting this back confirming that the update proceeds immediately.

PR #331, opened 2025-11-25
validation of user vacation request.

Dshad-Anon:


src/utils/vacations-utils.ts
```diff
@@ -107,14 +115,15 @@ export const validateUserVacationRequest = (
   vacationRequestData: { days?: number | string | null },
   currentYear: string,
   setError: (msg: string) => void,
-  setLoading: (loading: boolean) => void
+  setLoading: (loading: boolean) => void,
+  options?: {isAdmin?: boolean}
```
Jdallos: This seems a little strange to add this as options and then an object containing isAdmin, why not just pass the isAdminBoolean- would be clearer.

src/utils/vacations-utils.ts
```diff
@@ -107,14 +115,15 @@ export const validateUserVacationRequest = (
...
 ): boolean => {
   const unspentDays = Number(
     getDays(userToValidate?.attributes?.unspentVacationDaysByYear, currentYear)
   );
   const requestedDays = Number(vacationRequestData.days ?? 0);
 
-  const { valid, errorMessage } = validateVacationRequestDays(requestedDays, unspentDays);
+  const { valid, errorMessage } = validateVacationRequestDays(requestedDays, unspentDays,options);
```
Jdallos:
```suggestion
  const { valid, errorMessage } = validateVacationRequestDays(requestedDays, unspentDays, options);
```

src/utils/vacations-utils.ts
```diff
@@ -76,22 +76,30 @@ export function formatVacationDaysPayload(vacationDays: VacationDays): VacationD
  */
 export const validateVacationRequestDays = (
   requestedDays: number,
-  unspentDays: number
+  unspentDays: number,
+  options?: { isAdmin?: boolean }
```
Jdallos: This seems a little strange to add this as options and then an object containing isAdmin, why not just pass the isAdminBoolean- would be clearer.

src/components/screens/vacation-requests-screen.tsx
```diff
@@ -363,6 +364,32 @@ const VacationRequestsScreen = () => {
...
+
+          const selectedUser = await usersApi.findUser({ userId: vacationRequest.userId });
+
+          const isValid = validateUserVacationRequest(
+            selectedUser,
+            vacationRequest,
+            currentYear,
+            setError,
+            setLoading,
+            { isAdmin: adminMode }
```
Jdallos: why not pass admin mode (see later comments)

src/utils/vacations-utils.ts
```diff
@@ -76,22 +76,30 @@ export function formatVacationDaysPayload(vacationDays: VacationDays): VacationD
...
 ): { valid: boolean; errorMessage?: string } => {
   if (requestedDays > unspentDays) {
     const formatString = (template: string, values: Record<string, string | number>) =>
       template.replace(/\{(\w+)\}/g, (_, key) => String(values[key] ?? ""));
```
Jdallos: We don't need the replace or the regEx to do this, rather we can do it directly using the strings localization library- there should be examples of passing a dynamic value to a string in the code already

PR #358, opened 2025-12-17
Software application page is now showing username and also backbutton is consistent with rest of app in bottom.

Dshad-Anon:


MariaKatariinaH (approved):
Just noticed that the creation date is mm/dd/yyyy and should probably be changed to be dd/mm/yyyy.

PR #359, opened 2025-12-17
Created onboarding info to the wiki documentation screen

MariaKatariinaH:
Implementation copied from the home screen onboarding to be consistent. Id is used as a selector in the onboarding element, but Biome is warning if hardcoded ids are used.

### DIFF
--- a/src/components/contexts/delete-confirmation-dialog.tsx
+++ b/src/components/contexts/delete-confirmation-dialog.tsx
@@ -1,5 +1,6 @@
 import { Divider, Typography } from "@mui/material";
 import strings from "src/localization/strings";
+import type { DeleteItemType } from "src/types";
 import GenericDialog from "../generics/generic-dialog";
 
 /**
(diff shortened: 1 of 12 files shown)


src/components/onboarding/Onboarding.tsx
```diff
@@ -13,9 +21,9 @@ const POPUP_HEIGHT = 140;
  *
  * Displays guided onboarding tooltips around selected UI elements.
  */
-const Onboarding: React.FC = () => {
+const Onboarding: React.FC<OnboardingProps> = ({ screen }) => {
```
Alexander-K-byte:
```suggestion
const Onboarding = ({ screen }: OnboardingProps) => { ... }
```
Recommended not to use React.FC any longer, and I can see you were just adding props to the line that was already there but this is as good a time as any to refactor it

Alexander-K-byte (approved):
Old React.FunctionComponent should be changed as suggested

src/components/onboarding/Onboarding.tsx
MariaKatariinaH: Replaced React.FC as suggested.

Dshad-Anon (approved):

PR #371, opened 2025-12-30
Fixed some finnish typos

MariaKatariinaH:
There might be some adjustments left, but some obvious typos fixed.


Alexander-K-byte (approved):
Nothing much I can review since Finnish is not my native language.

PR #375, opened 2025-12-31
wiki-bugs

Dshad-Anon:


src/localization/en.json
```diff
@@ -443,9 +443,10 @@
     "labelTitle": "Title",
     "labelPath": "Path",
     "labelTags": "Tags",
-    "labelImage": "Image URL",
-    "labelDescription": "Desciption",
+    "labelImage": "Image Link",
```
Alexander-K-byte: Why was this changed from URL?  A link is normally something that will redirect you, URL is the web address.

src/localization/fi.json
```diff
@@ -443,9 +443,10 @@
     "labelTitle": "Otsikko",
     "labelPath": "Polku",
     "labelTags": "Tunnisteet",
-    "labelImage": "Kuvan URL-osoite",
+    "labelImage": "Kuvan linkki",
```
Jdallos: Same here as for english locale

PR #397, opened 2026-02-16
Fixed /balance, /sprintview access without opting in.

### LINKED ISSUE #396: Can access balance and sprintview pages without opting in
Can access balance and sprintview pages without opting in by typing in url bar.

MikaelPL1:


src/components/generics/opt-in-redirect.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      navigate("/");
+    }, 5000);
+
+    return () => clearTimeout(timer);
+  }, [navigate]);
+
+  return <ErrorScreen title="Not Opted In" message="You are not opted in. Redirecting to home page..." />;
```
Dshad-Anon: Localize the error title and message.See other example in codebase.

Dshad-Anon (approved):

rashinikarunarathne (approved):

src/components/generics/opt-in-redirect.tsx
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from "react";
+import { useNavigate } from "react-router-dom";
+import NotOptedInScreen from "../screens/not-opted-in-screen";
+
+const OptInRedirect = () => {
```
Jdallos: All components and functions should have TSDocs they describe the function/ component, and detail any parameters used.
So a common comment you might see from me in the future is missing TSDocs.

here is an example for a component like here:
/**
 * Balance screen component.
 */

And here is a typical function example:
/**
 * Check if a user is opted in (has a real Severa user ID, without dev fallbacks)
 *
 * @param user
 */

Jdallos (changes requested):

PR #420, opened 2026-03-19
Edit and delete buttons now showing properly 

arodriguezvaz:
Screenshot from 2026-03-19 09-31-00

The edit button was not shown entirely so I changed the row sizes. I  also realize that there was a delete button that was not showing unless you resize the column, so I changed the column default width.

### DIFF
--- a/src/components/questionnaire/questionnaire-table.tsx
+++ b/src/components/questionnaire/questionnaire-table.tsx
@@ -488,7 +488,7 @@ const QuestionnaireTable = () => {
 
         <DataGrid
           ref={dataGridRef}
-          getRowHeight={() => "auto"}
+          rowHeight={60}
           sx={{
             margin: 0,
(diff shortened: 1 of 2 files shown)


Dshad-Anon (approved):

MikaelPL1 (approved):

rashinikarunarathne (approved):

src/components/questionnaire/questionnaire-table.tsx
```diff
@@ -20,6 +20,7 @@ import {
 } from "@mui/material";
 import { DataGrid, type GridRenderCellParams, type GridRowParams } from "@mui/x-data-grid";
 import { useAtomValue, useSetAtom } from "jotai";
+import { min } from "lodash";
```
Jdallos: Why is this import here, it seems to be unused and should be removed?

Jdallos (approved):

PR #428, opened 2026-04-01
integrated user status endpoint with front-end

rashinikarunarathne:


petrasoininen (approved):

arodriguezvaz (approved):

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -24,254 +25,316 @@ import { useLambdasApi } from "src/hooks/use-api";
 import strings from "src/localization/strings";
 import BackButton from "../generics/back-button";
 
+const getUserId = (user: UserFlextime["user"]): string | undefined => {
```
Jdallos: Missing TSDOcs

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -24,254 +25,316 @@ import { useLambdasApi } from "src/hooks/use-api";
...
+	return (user as any).id;
+};
```
Jdallos, line 36:
```suggestion
};

```

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -60,11 +64,40 @@ const EmployeeFlextimeScreen = () => {
...
+                  ...u.user,
+                  attributes: {
+                    ...u.user.attributes,
+                    isActive: active
+                  }
+                }
+              }
+            : u
+        )
+      );
```
Jdallos, line 109: This code is quite difficult to read due to the nested conditional and the indentations due to dense object update.
In cases like this a good start point would be to make things as clear as possible, so rename u to user. We could also get rid of the ternary itself by explicitly checking. SO both of those in place  would look like:

 setUsersFlextime((prev) =>
      prev.map((item) => {
        if (getUserId(item.user) !== userId) {
          return item;
        }

        return {
          ...item,
          user: {
            ...item.user,
            attributes: {
              ...item.user.attributes,
              isActive: active
            }
          }
        };
      })
    );


That's quite reasonable in this case. It's bordering on whether to extract the update into it's own function and call it. I think it's fine with the above change but extracting the function would look like this:

const updateUserActiveState = (
  userFlextime: typeof usersFlextime[number],
  userId: string,
  active: boolean
) => {
  if (getUserId(userFlextime.user) !== userId) {
    return userFlextime;
  }

  return {
    ...userFlextime,
    user: {
      ...userFlextime.user,
      attributes: {
        ...userFlextime.user.attributes,
        isActive: active
      }
    }
  };
};


const handleStatusChange = async (userId: string, active: boolean) => {
  if (!usersApi) return;

  try {
    await usersApi.updateUserStatus({
      userId,
      updateUserStatusRequest: {
        isActive: active
      }
    });

    setUsersFlextime((prev) =>
      prev.map((userFlextime) =>
        updateUserActiveState(userFlextime, userId, active)
      )
    );
  } catch {
    setError("Failed to update user status");
    loadFlextimeData();
  }
};

Choose whichever seems clearest and best to you. Remember to add TSDOcs

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -60,11 +64,40 @@ const EmployeeFlextimeScreen = () => {
     }
   };
 
+  const handleStatusChange = async (userId: string, active: boolean) => {
```
Jdallos, line 81: TSDOcs missing

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -60,11 +64,40 @@ const EmployeeFlextimeScreen = () => {
...
+                    isActive: active
+                  }
+                }
+              }
+            : u
+        )
+      );
+    } catch {
+      setError("Failed to update user status");
+      loadFlextimeData();
```
Jdallos: Should this be awaited?

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -198,73 +231,82 @@ const EmployeeFlextimeScreen = () => {
...
+                        </Typography>
+                      </TableCell>
+                      <TableCell align="center">
+                        <Select
+                          value={isActive ? "active" : "inactive"}
+                          onChange={(e) => {
+                            const userId = getUserId(userData.user);
+                            if (!userId) return;
+
+                            handleStatusChange(userId, e.target.value === "active");
```
Jdallos, line 298: Again this is an async function so shouldn't we await here?

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -60,11 +64,40 @@ const EmployeeFlextimeScreen = () => {
     }
   };
 
+  const handleStatusChange = async (userId: string, active: boolean) => {
+    try {
+      if (!usersApi) return;
+      await usersApi.updateUserStatus({
```
Jdallos, line 87: Typically if we have an async function we have some kind of loading spinner or state that is activated while we are waiting for the response. DO you think that should be added here? If so be consistant with other loading spinners/ progress indicators used for similar features in the project.

Jdallos (changes requested):

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -24,13 +25,23 @@ import { useLambdasApi } from "src/hooks/use-api";
 import strings from "src/localization/strings";
 import BackButton from "../generics/back-button";
 
+/**
+ * Extracts the user d from a UserFlextime user object.
+ * @param user- The user object from UserFlextime
+ * @returns The user ID if available
+ */
+const getUserId = (user: UserFlextime["user"]): string | undefined => {
+  return (user as any).id;
```
Jdallos: We are typing the user in the function definition and it should have the id property on it, so I'm not sure why you have the as any casting here. Can it be removed without causing some error?

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -59,12 +70,34 @@ const EmployeeFlextimeScreen = () => {
       setLoading(false);
     }
   };
-
   /**
-   * Formats a flextime balance in hours with appropriate sign.
-   * @param hours - The number of hours.
+   * update the active status of a user.
```
Jdallos:
```suggestion
   * Update the active status of a user.
   * 
```

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -59,12 +70,34 @@ const EmployeeFlextimeScreen = () => {
...
+   * @param userId The ID of the user to update
+   * @param active Whether the user should be active or inactive
+   * @returns
```
Jdallos:
```suggestion
```

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -59,12 +70,34 @@ const EmployeeFlextimeScreen = () => {
...
+   */
+  const handleStatusChange = async (userId: string, active: boolean) => {
+    try {
+      if (!usersApi) return;
```
Jdallos, line 83:
```suggestion
      if (!usersApi) return;
      
```

Jdallos: Line break after return statements improves readability

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -59,12 +70,34 @@ const EmployeeFlextimeScreen = () => {
...
+      setUpdatingUserId(userId);
+      await usersApi.updateUserStatus({
+        userId,
+        updateUserStatusRequest: {
+          isActive: active
+        }
+      });
+      await loadFlextimeData();
```
Jdallos: To clarify we are loading flextime data becuase the isActive status is included in that data? If so I can see the spec is also calling it flextimedata:
resourceAllocationsApi.listUsersFlextime()
But i wonder if we might want to rename it with a more appropriate name. Your call but this could be a small seperate task, if you could decide on if it's worthwhile and document please.

Jdallos: Also test the UX of loadingFlextimeData, if for example updating a single users status sets the loading state, but this seems to only affect the getTotalBalance component, not the main component render, so do we see any loader for the status update, do we need to? Check loading UX with our designer.

Also at minimum if the updating of a status causes the unrelated balance to flicker with a loading state that seems a bit strange, might want to think how to improve the loading handling for a clean UX.

src/components/screens/employee-flextime-screen.tsx
```diff
@@ -198,73 +231,67 @@ const EmployeeFlextimeScreen = () => {
...
-                                borderColor: theme.palette.action.disabled,
-                                color: theme.palette.text.disabled
-                              }
-                        }
-                      />
-                    </TableCell>
-                  </TableRow>
-                ))}
+                      </TableCell>
+
```
Jdallos:
```suggestion
```

Jdallos (approved):

src/components/screens/employee-flextime-screen.tsx
rashinikarunarathne: removed the unnecessary as any and updated typing accordingly.

src/components/screens/employee-flextime-screen.tsx
rashinikarunarathne: i removed full reload and now update the affected user locally to avoid unnecessary loading and UI flicker.

PR #432, opened 2026-04-08
software registry admin edit application bug fixed

rashinikarunarathne:


arodriguezvaz (approved):

petrasoininen (approved):

src/components/software-registry/AddSoftwareModal.tsx
```diff
@@ -64,7 +64,7 @@ const AddSoftwareModal = ({
   const { usersApi } = useLambdasApi();
   const setError = useSetAtom(errorAtom);
   const [userList, setUserList] = useState<User[]>([]);
-  const [software, setSoftware] = useState<SoftwareRegistry>(softwareData || initialSoftwareState);
+  const [software, setSoftware] = useState(initialSoftwareState);
```
Jdallos, line 67: It looks like TS is inferrring the type here so the <SoftwareRegistry> isn't strictly necessary. In this context it would improve readability IMO so maybe add ti back in?

src/components/software-registry/AddSoftwareModal.tsx
```diff
@@ -103,7 +111,9 @@ const AddSoftwareModal = ({
    * Handle form field reset to initial values.
    */
   const resetForm = () => {
-    setSoftware(initialSoftwareState);
+    if (!softwareData) {
```
Jdallos, line 114: So is this check added if we use the softwareModal to update an existing software also?
If so then this is fine but I'd then rename the file and the component (any the AddSoftwareModalProps type) to SoftwareModal (or softwareDialog).

If we are not using this for updating, then please explain why this condition was added when resetting the form.

src/components/software-registry/AddSoftwareModal.tsx
```diff
@@ -184,14 +194,17 @@ const AddSoftwareModal = ({
             <CloseIcon />
           </IconButton>
           <Typography variant="h6" marginBottom={4}>
-            {strings.softwareRegistry.addSoftware}
+            {softwareData
+              ? strings.softwareRegistry.editApplication
+              : strings.softwareRegistry.addApplication}
```
Jdallos, line 199: I guess the answer to my previous comment is yes, this is being used for updating. Thats good, please rename the components file and type as suggested previously

Jdallos (approved):

PR #449, opened 2026-06-03
Feature 442 duplicate vacation requests and admin dates change

NeonValtokari:
- Admin is now able to change the _days consumed_ of vacation without the dates changing
- Admin is now able to change the dates of vacation and the days consumed change accordingly
- Default days of vacation are NOT shown in the vacation request table, instead it shows days consumed
- When creating or drafting a vacation request employee can no longer create multiple identical ones by accidentally or on purpose spamming the submit button

Fixes #442

### DIFF
--- a/src/components/vacation-requests-table/vacation-requests-table-toolbar/toolbar-form/toolbar-form-fields.tsx
+++ b/src/components/vacation-requests-table/vacation-requests-table-toolbar/toolbar-form/toolbar-form-fields.tsx
@@ -93,7 +93,7 @@ const ToolbarFormFields = ({
     originalStartDateRef.current = null;
   }, [toolbarFormMode]);
 
-  // Update vacation request whenever date range changesei
+  // Update vacation request whenever date range changes
   useEffect(() => {
     if (!dateRange.start || !dateRange.end) return;
     // Store original end date on first load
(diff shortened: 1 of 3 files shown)


rashinikarunarathne (approved):

arodriguezvaz (approved):

src/components/vacation-requests-table/vacation-requests-table-toolbar/toolbar-form/toolbar-form-fields.tsx
```diff
@@ -93,17 +90,26 @@ const ToolbarFormFields = ({
    */
   useEffect(() => {
     originalEndDateRef.current = null;
+    originalStartDateRef.current = null;
   }, [toolbarFormMode]);
 
-  // Update vacation request whenever date range changes
+  // Update vacation request whenever date range changesei
```
Jdallos: typo?

src/components/vacation-requests-table/vacation-requests-table-toolbar/toolbar-form/toolbar-form-fields.tsx
```diff
@@ -113,11 +119,14 @@ const ToolbarFormFields = ({
         days
       });
     } else {
-      setVacationRequestData({
-        ...vacationRequestData,
-        startDate: dateRange.start.toJSDate(),
-        endDate: dateRange.end.toJSDate()
-      });
+      if (isModified) {
```
Jdallos: SonarCloud analysis comment seems legit. I don't see why we need if in the else block (anymore) so it can als obe removed

src/utils/time-utils.ts
```diff
@@ -104,6 +104,9 @@ export const calculateTotalVacationDays = (
 };
 
 /**
+ * NOTE: This function is no longer in use, should it be removed, or is it
```
Jdallos: Fair point.
Would it be reasonable to keep this but rather use it when creating a vacation. So rather than selecting the dates, a user (not admin) could also select the number of days and the end date would be filled in appropriately.
Up to you, feel free to check with designer. If yes, create the trello card and feel free to self allocate, or remove the function as originally commented.

Jdallos (approved):

PR #450, opened 2026-06-03
Merge ui-redesign branch with develop

NeonValtokari:


rashinikarunarathne (approved):

src/theme.tsx
```diff
@@ -5,71 +5,161 @@ import PoppinsMedium from "../resources/fonts/poppins/Poppins-Medium.ttf";
...
+      bodySmall: {
+        fontSize: 14,
+        lineHeight: "160%"
       },
-      body2: {
-        fontWeight: 400,
-        fontSize: 16
+      caption: {
+        fontSize: 12,
+        lineHeight: "150%"
```
Jdallos, line 158: Typically in projects we use line height as unitless number, so 150% would be 1.5.
I'd suggest making the change throughout the theme if we want to start with the correct pattern.

However if the figma itself is expressing line heights as percentage then leave as it.

src/theme.tsx
```diff
@@ -117,6 +207,26 @@ export const createAppTheme = (mode: "light" | "dark"): Theme =>
       MuiTooltip: {
         defaultProps: {
           arrow: true
+        },
+
```
Jdallos:
```suggestion
```

src/theme.tsx
```diff
@@ -117,6 +207,26 @@ export const createAppTheme = (mode: "light" | "dark"): Theme =>
...
+            padding: `${theme.spaces.s} ${theme.spaces.s}`,
+            display: "flex",
+            flexDirection: "column",
+            justifyContent: "center",
+            alignItems: "flex-start",
+            gap: theme.spaces.s,
+            ...theme.typography.caption,
+            fontWeight: 400
+          }),
+
```
Jdallos:
```suggestion
```

src/theme.tsx
Jdallos: Unnecessary line break

src/theme.tsx
```diff
@@ -171,15 +270,27 @@ export const createAppTheme = (mode: "light" | "dark"): Theme =>
     }
   });
 
+/**
+ * Color tokens for wiki screen UI.
+ *
+ * These values have not yet been updated to match UI changes
+ * and need further review.
```
Jdallos, line 276: Looks like it should be marked as TODO:

src/theme.tsx
```diff
@@ -171,15 +270,27 @@ export const createAppTheme = (mode: "light" | "dark"): Theme =>
...
     text: theme.palette.text.primary,
     border: theme.palette.divider
   }
 });
 
+/**
+ * Custom theme tokens and styles.
+ *
+ * These are yet to be aligned with UI changes and should
+ * be revisited.
```
Jdallos, line 292: Looks like TODO:

Jdallos (approved):

PR #461, opened 2026-07-01
Feature 454 homepage header redesign

Eelis96:
Update header language select, user dropdown, background color, and home page greeting text according to design.


src/utils/user-utils.ts
```diff
@@ -18,3 +18,27 @@ export const getSeveraUserId = (user: User | undefined): string => {
...
+ */
+export const getDisplayName = (user?: User): string => {
+  if (user?.firstName) {
+    return user.firstName;
+  }
+
+  if (user?.email) {
+    const localPart = user.email.split("@")[0];
+    const withoutPrefix = localPart.replace(/^ext-/i, "");
+    const firstPart = withoutPrefix.split(/[._-]/)[0];
```
NeonValtokari: Problematic for people with two part names. For example "Jarmo-Hemuli" would become just "Jarmo" due to splitting on "-" and taking only the first segment.

NeonValtokari (approved):

rashinikarunarathne (approved):

PR #477, opened 2026-07-16
Feature 476 on call and vacation card for new UI

NeonValtokari:
- On-call and vacation card for new UI
- Vacations card is now same for all users

Closes #476


Eelis96 (approved):

src/components/vacation-requests-table/vacationStatusIndicator.tsx
```diff
@@ -17,9 +18,10 @@ interface StatusIndicatorProps {
  * of the indicator based on the status
  * @returns Themed MUI ToolTip component
  */
-const VacationStatusIndicator = ({ statusColor }: StatusIndicatorProps) => {
+const VacationStatusIndicator = ({ statusColor, disabled }: StatusIndicatorProps) => {
   const theme = useTheme();
-  const colors = getBadgeColor(statusColor, theme);
+  const color = disabled || statusColor;
```
Jdallos: In badges.tsx you have disabled prop we pass here, which is either "disabled" or status. status is also what statusColor here is.
So isn't const color = disabled || statusColor; redundant as disabled will always exist? so the || statusColor is never truthy.

So if I'm correct you can remove statusColor prop entirely and rename disabled to something more appropritate e.g. statusColor- assuming "disabled" is in fact a color. Although even that name in this context is a little confusing as we are using it in an animation. So perhaps status is the best name?

src/components/generics/badges.tsx
```diff
@@ -92,11 +93,14 @@ type PillBadgeProps =
...
  */
-export const PillBadge = ({ variant, status, children }: PillBadgeProps) => {
+export const PillBadge = ({ variant, status, children, hidden }: PillBadgeProps) => {
   const theme = useTheme();
+  const color = hidden ? "disabled" : status;
 
   const approvalIcon =
-    variant === "approvalBadge" ? <VacationStatusIndicator statusColor={status} /> : undefined;
+    variant === "approvalBadge" ? (
+      <VacationStatusIndicator statusColor={status} disabled={color} />
```
Jdallos: See note in vacationStatusIndicator.tsx about how I suspect you only need one prop here

src/components/home/oncall-card.tsx
```diff
@@ -36,37 +47,33 @@ const OnCallCard = () => {
...
-   *
-   * @returns JSX.Element containing current on-call person
-   */
-  const getCurrentOnCallPerson = () => {
+  const renderOnCallCard = () => {
     const currentWeek = DateTime.now().weekNumber;
-    const currentOnCallPerson = onCallData.find(
-      (item) => Number(item.week) === currentWeek
-    )?.username;
+    const currentOnCallPerson = onCallData.find((item) => Number(item.week) === currentWeek)?.email;
```
Jdallos:
```suggestion
    const currentOnCallPersonEmail = onCallData.find((item) => Number(item.week) === currentWeek)?.email;
```

src/components/home/vacations-card.tsx
```diff
@@ -172,99 +146,69 @@ const VacationsCard = () => {
...
-                    </Typography>
-                    <Grid container>
-                      {vacationInfoListItems.map((vacationInfoListItem, index) =>
-                        renderVacationInfoItem(vacationInfoListItem, index)
-                      )}
-                    </Grid>
-                  </>
+        <Box>
+          {earliestUpcomingVacationRequest &&
+            DateTime.fromJSDate(earliestUpcomingVacationRequest.startDate) > DateTime.now() && (
```
Jdallos: This could be made clearer by combining these conditions into a named variable. SO before this render block you would have something like:

const earliestUpcomingVacationRequest = earliestUpcomingVacationRequest &&
            DateTime.fromJSDate(earliestUpcomingVacationRequest.startDate) > DateTime.now();

Please confirm yourself if the name reflects the logic accurately.

src/utils/user-name-utils.ts
```diff
@@ -17,65 +17,54 @@ export const getDisplayName = (user?: User): string => {
...
-      .map((part) => capitalize(part))
-      .join("-");
+    const firstPart = parseFromEmail(user.email)[0];
+    return firstPart;
   }
 
   return "";
 };
 
+export const parseFromEmail = (email: string) => {
```
Jdallos: Missing TSDOcs

Jdallos (approved):

